PR #5, opened 2019-12-26
do not addRef on implicit objects

wwanderley:
After some investigation, I figured out that calling PL/SQL Stored Procedures may cause some memory leaks and SIGSEGV:

- Some objects (like pl/sql collections and attributes as complex types) don't need to call dpiObject_addRef; See https://github.com/oracle/odpi/issues/122

- Changed ObjectType.close() to avoid "signal SIGSEGV: segmentation violation"

### DIFF
--- a/data.go
+++ b/data.go
@@ -25,6 +25,7 @@ import (
 type Data struct {
 	ObjectType    ObjectType
 	dpiData       C.dpiData
+	implicitObj   bool
 	NativeTypeNum C.dpiNativeTypeNum
 }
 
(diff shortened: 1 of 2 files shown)


tgulacsi:
And some tests would be great for such reference counting things - this is not the first or second time when we try to squash such bgs.

Maybe even such an awkward one that spawns a subtest with DPI_DEBUG_LEVEL=2 and analyze the printed logs.

I can help with such machinery, but please help with the concrete SIGSEGV and memory leaking code!

Thanks,
Tamás

Issue #22, opened 2020-02-22
CQN subscription callback not working in Docker


dinvlad:
**Describe the bug** [this might be a feature request after all, please see the next post]
When running inside Docker, CQN subscription query registration seems to crash when the first change is triggered (but before the callback has a chance to run).

**To Reproduce**
1. Register a QRCN query according to https://github.com/godror/godror/blob/master/z_qrcn_test.go
2. Build a Docker image using
```Dockerfile
FROM golang AS build
WORKDIR /build
COPY go.* *.go ./
RUN go build -o main

FROM oraclelinux:7-slim
RUN yum -y -q install oracle-release-el7 && \
    yum-config-manager --enable ol7_oracle_instantclient >/dev/null && \
    yum -y -q install oracle-instantclient19.5-basic
COPY --from=build /build/main /
ENTRYPOINT ["/main"]
```
3. Run the Docker image.
4. Observe that the subscription query is registered successfully (no errors returned):
```go
func printReg() {
	var regID, tableName string
	row := testDb.QueryRow("SELECT regid, table_name FROM USER_CHANGE_NOTIFICATION_REGS")
	row.Scan(&regID, &tableName)
	log.Println(regID, tableName)
}()
testDb.Exec("INSERT INTO test_subscr (i) VALUES (1)")
printReg()
```
5. Observe that registration params were printed before the insert, and are missing after. No events are logged either, despite having a log statement in the callback.
6. Compile and run the same code on **macOS** or **RHEL** (without Docker), observe that the events are received and the registration survives.

**Expected behavior**
Insertion event should be logged, and the registration should remain alive after the event is triggered.

**Screenshots**

**Your oracle client version**
19.5 (same issue with 18.3)

**Your godror version**
v0.9.1

**Your go version**
1.13.8

**Your gcc version**
8.3.0

**Machine (please complete the following information):**
 - OS: Oracle Linux 7 (same behavior on Debian) in a **Docker** on **macOS 10.14.6** and **Amazon Linux (AWS Fargate instance)**
 - Architecture: x86_64

**Additional context**
As mentioned, I tried other base images (Debian) but the same behavior is observed.

Additionally, I tried `CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build` but that didn't help.

Finally, I tried both `strace` and `DPI_DEBUG_LEVEL=100`, but they don't reveal any messages/calls after the change is triggered (or maybe I didn't run those with proper options?).

[...]

PR #23, opened 2020-02-22
Add NewSubscriptionWithParams() method for extra params

dinvlad:
This PR addresses https://github.com/godror/godror/issues/22, by adding `NewSubscriptionWithParams()` method. It accepts a new `SubscriptionParams` struct that provides `ipAddress`, `portNumber` and `clientInitiated` options for `dpiSubscrCreateParams`.

This struct can also be expanded in the future to include even more subscription options.

### DIFF
--- a/orahlp.go
+++ b/orahlp.go
@@ -479,7 +479,7 @@ type Conn interface {
 	ClientVersion() (VersionInfo, error)
 	ServerVersion() (VersionInfo, error)
 	GetObjectType(name string) (ObjectType, error)
-	NewSubscriptionWithParams(SubscriptionParams, func(Event)) (*Subscription, error)
+	NewSubscriptionWithParams(SubscriptionParams) (*Subscription, error)
 	NewSubscription(string, func(Event)) (*Subscription, error)
 	Startup(StartupMode) error
 	Shutdown(ShutdownMode) error
--- a/orahlp.go
+++ b/orahlp.go
@@ -479,6 +479,7 @@ type Conn interface {
 	ClientVersion() (VersionInfo, error)
 	ServerVersion() (VersionInfo, error)
 	GetObjectType(name string) (ObjectType, error)
+	NewSubscriptionWithParams(SubscriptionParams, func(Event)) (*Subscription, error)
 	NewSubscription(string, func(Event)) (*Subscription, error)
 	Startup(StartupMode) error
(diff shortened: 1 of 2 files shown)


tgulacsi:
Nice, thank you for it!

Though I have a bigger question: as godror (intentionally) does not have a stable (v1.0) version yet, such (NewSubscription -> NewSubscriptionWithParams) backwards-incompatible modifications are allowed.

So, @dinvlad, @cjbj, @anthony-tunininga, do we want to tote the ballast of this non-configurable NewSubscription beside the proper NewSubscriptionWithParam, or just replace it?

I vote for replacing it - I don't know anybody does use this subscription feature.

A more backwards-compatible variant could be adding optional parameters:

    func NewSubscription(name string, callback func(Event)) (*Subscription, error),
                                        options ...SubscrOption)

with

```lang:go
    type SubscrOption func(*subscrParams)

// IP address on which the subscription listens to receive notifications,
	// for a server-initiated connection.
	//
	// The IP address can be an IPv4 address in dotted decimal format such as 192.1.2.34
	// or an IPv6 address in hexadecimal format such as 2001:0db8:0000:0000:0217:f2ff:fe4b:4ced.
	//
	// By default, an IP address will be selected by the Oracle client.	// Port number on which to receive notifications, for a server-initiated connection.
	// The default value of 0 means that a port number will be selected by the Oracle client.
    func SubscrHostPort(IPAddress	string, Port uint32) SubscrOption

	// Specifies whether a client or a server initiated connection should be created.
	//
	// This feature is only available when Oracle Client 19.4
	// and Oracle Database 19.4 or higher are being used.
	func SubscrClientInitiated(bool) SubscrOptions
```

dinvlad:
Yep, thought about adding subscription options too. I'd be OK with either choice, and the existing interface is really concise, if one doesn't need any extra options.

dinvlad:
I'd say that codifying `SubscrOption`s would take a bit more code though..

anthony-tuininga:
> So, @dinvlad, @cjbj, @anthony-tunininga, do we want to tote the ballast of this non-configurable NewSubscription beside the proper NewSubscriptionWithParam, or just replace it?

I'd agree with you that replacing is the better option if you can get away with it! :-)

tgulacsi:
I went with the options route - see v0.13.0

dinvlad:
Awesome!

Issue #28, opened 2020-03-13
Connections to heterogeneous pools don't respect usernames and passwords set by godror.ContextWithUserPassw


schaefed:
**Describe the bug**
When opening a connection to an heterogeneous pool with `godror`, later changes to the username and password with `godror.ContextWithUserPassw` are not respected, connection attemps error out with `ORA-01017: invalid username/password; logon denied`.
This used to work with `goracle` up to version 1.14.

**To Reproduce**
1. Open a connection without username and password
```go
	params := godror.ConnectionParams{
		Username: "",
		Password: "",
		SID:  "MYSID",
	}
	db, err := sqlx.Open("godror", params.StringWithPassword())
```
2. Provide username and password
```go
ctx := godror.ContextWithUserPassw(
		context.Background(),
		"username",
		"password",
		"",
	)
```
3. Ping the generated context
```go
err := db.PingContext(ctx)
```

**Expected behavior**
A successful ping.

**Your oracle client version**
18.4.0.0.0

**Your godror version**
latest commit

**Your go version**
1.14

**Your gcc version**
e.g. 9.2.1

**Machine (please complete the following information):**
 - OS: ArchLinux
 - Architecture:  x86_64
 - Version: up to date

tgulacsi:
Can you help creating a test case?

PR #29, opened 2020-03-13
Add username and password to context before the connection attempt

schaefed:
A (too simple?) fix for #28 . Not sure if the mutation of `ConnectionParams` at the connection level has any unwanted side effects.


tgulacsi:
As neither the connector nor connector.ConnectionParams are pointers, the modification is local.
But I'll make this explicit.

PR #31, opened 2020-03-18
Sharding proof of concept

anthony-tuininga:
This PR is intended as a proof of concept for adding support for sharding. You will note it adds a NewConnector2() function as additional parameters are required. Ideally this should be integrated into the existing NewConnector() function if possible. There should also be a consolidation of the code and structures required to create a pool and acquire connections from it. Suggestions on the best way to do so is welcome. I can make any changes you suggest or am willing to review changes that you would like to make if you prefer that approach.

As you are likely aware, when acquiring connections from an Oracle pool, additional parameters may be specified. These include a username and pasword (for proxy authentication), a purity flag (for specifying whether an existing session or a new session should be acquired), a tag and "match any" flag (for session tagging), and sharding and super sharding keys (for sharding). This PR only addresses the sharding keys as I wanted to get feedback on the approach before going any further with this. Using any of these parameters would require the connector approach as the Go SQL package doesn't allow for these additional parameters.

All feedback welcome!


drv.go
```diff
@@ -137,6 +137,174 @@ type connPool struct {
...
+
+    // determine key to use for pool
+    poolKey := fmt.Sprintf("%s\t%s\t%s\t%d\t%d\t%d\t%s\t%s\t%s\t%t\t%t\t",
+            P.UserName, P.Password, P.DSN, P.MinSessions, P.MaxSessions,
+            P.SessionIncrement, P.WaitTimeout, P.MaxLifeTime, P.SessionTimeout,
+            P.Heterogeneous, P.EnableEvents)
+    fmt.Printf("Pool key: |%s|\n", poolKey)
+
+    // if pool already exists, return it immediately
+    if pool, ok := d.pools[poolKey]; ok {
```
tgulacsi: Serialize access with d.mu.Lock() and d.mu.Unlock():
d.mu.Lock()
pool, ok := d.pools[poolKey]
d.mu.Unlock()
if ok {

drv.go
```diff
@@ -137,6 +137,174 @@ type connPool struct {
...
 }
 
+func (d *drv) openPool(P *PoolParams) (*connPool, error) {
+
+    // determine key to use for pool
+    poolKey := fmt.Sprintf("%s\t%s\t%s\t%d\t%d\t%d\t%s\t%s\t%s\t%t\t%t\t",
+            P.UserName, P.Password, P.DSN, P.MinSessions, P.MaxSessions,
+            P.SessionIncrement, P.WaitTimeout, P.MaxLifeTime, P.SessionTimeout,
+            P.Heterogeneous, P.EnableEvents)
+    fmt.Printf("Pool key: |%s|\n", poolKey)
```
tgulacsi: if Log != nil {
    Log("pool key", poolKey)
}

drv.go
```diff
@@ -927,6 +1109,17 @@ func (d *drv) OpenConnector(name string) (driver.Connector, error) {
 // The returned connection is only used by one goroutine at a
 // time.
 func (c connector) Connect(context.Context) (driver.Conn, error) {
+    if c.PoolParams != nil {
+        pool, err := c.drv.openPool(c.PoolParams)
+        if err != nil {
+            return nil, err
+        }
+        connection := conn{drv: c.drv, timeZone: time.Local}
```
tgulacsi: Do we really know the server time zone here?

drv.go
```diff
@@ -137,6 +137,174 @@ type connPool struct {
...
+    }
+    C.dpiPool_setStmtCacheSize(dp, 40)
+    d.mu.Lock()
+    d.pools[poolKey] = &connPool{dpiPool: dp}
+    d.mu.Unlock()
+    return d.pools[poolKey], nil
+}
+
+
+func (c *conn) acquireFromPool(pool *connPool, P *ConnParams) error {
```
tgulacsi: Why not a method on *connPool ?
Seems a little bit strange to declare, half-initialize a connection, then acquire it from the pool.

drv.go
anthony-tuininga: Good point. I'll make that change.

drv.go
anthony-tuininga: I'll make this change, too.

drv.go
anthony-tuininga: I was simply following the code found in openConn() as seen [here](https://github.com/godror/godror/blob/master/drv.go#L190). Do you have a suggestion on what should be done instead?

drv.go
anthony-tuininga: That could be done. The issue is that there is a method already called acquireConn() which accepts ConnectionParams, and this does not include anything to deal with sharding. I created a new structure, but ideally there should only be one. That's a bit more invasive, however, so direction would be appreciated. I'm willing to make the changes if I know what sort you would like to see.

drv.go
tgulacsi: Ok, then leave it as is.

drv.go
tgulacsi: Yes, call it acquireConn2 for now - I can't see the whole picture yet, so leave the code merge for the future.

anthony-tuininga: Ok. I'll take care of that early next week.

anthony-tuininga: Method renamed as requested. I'll look into the merging of the two methods this week.

anthony-tuininga:
@tgulacsi, while going through the code for my changes I noted a comment about adding a value to a context via `godror.ContextWithUserPassw()`. This approach allows arbitrary values to be attached to a context, which can then be used with `db.ExecContext()`, `db.QueryContext()` and so forth. So there is a choice that needs to be made:

1. Do we create a new method `godror.ContextWithConnParams()` (or something equivalent) that contains all of the pooled connection acquisition parameters. This would include not just user name, password, connection class, but also sharding keys and the various tagging parameters that are available.

2. Do we continue along the path I have already started on? Have `godror.NewConnector()` accept pool parameters and connection parameters and then use `sql.OpenDB()`.

I have already made most of the changes needed to eliminate the duplicated code in my original set of changes -- but thought before I continue further I should ask your opinion on which approach makes more sense. The first approach requires the use of a modified context for all methods but permits you to use a single DB. The second approach requires multiple DBs but permits you to use the regular methods. Please advise! Thanks.

anthony-tuininga:
Some examples with both options.

The first option would look like this (error handling removed for brevity):

```
const dsn = "user/password@host/service"
const sql = "select sysdate from dual"
db, _ := sql.Open("'godror", dsn)
defer db.Close()
shardingKey := "A string"
params := godror.ConnParams{ShardingKey: []interface{}{shardingKey}}
ctx := godror.ContextWithConnParams(context.Background(), params)
row := db.QueryRowContext(ctx, sql)
```

The second option would look like this:

```
poolParams := godror.PoolParams{UserName: "user", Password: "password", DSN: "host/service"}
connParams := godror.ConnParams{ShardingKey: []interface{}{shardingKey}}
connector := godror.NewConnector2(&poolParams, &connParams)
db := sql.OpenDB(connector)
defer db.Close()
row := db.QueryRow(sql)
```

tgulacsi:
Using Context is always a hack. 
Even adding an Option (see godror.FetchRowCount) is more explicit, I think.

So I'd go with the second option - we can shovel the sharding keys into the connection string, too.
Or allow only the sharding key setup with the Context.

Anyway, passing params with Context should be for override, you should be able to set the defaults with the connection string / directly with ConnectionParams.

BTW, how is this shardingKey used?
https://docs.oracle.com/en/database/oracle/oracle-database/18/lnoci/oci-interface-for-using-shards.html#GUID-C64136C2-CF65-4507-8811-9A03946CC817
says that it selects a shard based on the data key - for me that suggests it is a per-connection property, not a per-pool property.

How often does this sharding key change? If it is a per-connection of per-statement thing, than it should not be part of the pool properties, only statement-level or Context-defined Option.

But I don't understand them yet.

anthony-tuininga:
Thanks, @tgulacsi. I'll continue on with what I was doing, then! It may even be useful to implement both approaches at some point but I'll start with option 2.

To give a bit of an explanation of sharding that will hopefully help you: a sharded database consists of two or more database instances that contain the same table structure (catalog). The sharding key is used to determine which of these database instances contains the data you are interested in querying or updating. As such it is indeed a connection property, not a pool property. You'll note in the code I have provided so far that it is found in the ConnParams structure and not the PoolParams structure. The sharding key is going to change as often as your application requires it. The geeneral approach is that (for example) a web application acquires a connection from the pool specifying the sharding key to ensure that it connects to the right instance, performs a number of queries or updates and then releases the connection back to the pool. This is done explicitly with other language libraries but Go assumes that there is no need to provide connection specific properties at any time other than when a DB is created -- thus the need for the connector (or context properties).

tgulacsi:
Thanks for the clarification!

That means it needs more thinking - Go's *sql.DB is a pool, and here we want to override the connection selection algorithm for sharding: we want to let Oracle return a pooled connection for the actual sharding key, and for the fresh one, not a previously selected one.

Here giving the sharding key by Context won't be enough: the connection is returned from the Go pool, and reaches the driver (and Oracle) only when no Go-pooled connection exist.

So the solution is to limit the cached Go connections by setting `db.SetMaxIdleConns(0)` - this will present a small performance loss as all connections must be recreated on Go side, but will allow Oracle to return a shard-specific connection for each shard key.

This should be documented, but will work. 
And you're right, the shard key will be a ConnParam, not a PoolParam.

anthony-tuininga:
I've just pushed changes to eliminate the duplicate code. There are still some things that should be addressed but they can happen afterwards -- assuming you are satisfied with these changes!

The new code has a number of new methods that replalce the ones that were there before with a bunch of comments in the header which hopefully explain their existence. If you have questions, let me know and I'll augment the comments as needed. A brief description follows:

- drv.initCommonCreateParams() -- initializes the common creation parameters used for creating pools and standalone connections

- drv.createConn()  -- creates an ODPI-C connection, either standalone or pooled

- drv.createConnFromParams() -- wrapper function which creates an ODPI-C connection using pool and connection parameters

- drv.getPool() -- either creates a new pool or uses an existing pool from the cache [NOTE: there was a bug before in that multiple pools could be created by multiple goroutines]

- drv.createPool() -- creates the ODPI-C pool

The original ConnectionParams structure is retained. It is used to hold the parameters specified in the string passed to sql.Open(). Internally the new PoolParams() and ConnParams() structures are used. You can see the mapping in drv.OpenConnector().

I did add godror.ContextWithParams() as a means of passing ConnParams directly without having to create a new connector (option 1 noted above) and changed godror.ContextWithUserPassw() to call this instead. So I ended up adding support for both options. :-)

I noted that some of the test suite was internally creating multiple pools unintentionally and therefore the test was "passing" whereas now it does not. I can create a separate change for the test suite or include it with this one if you prefer.

Further comments welcome!

drv.go
```diff
@@ -166,317 +168,405 @@ func (d *drv) init() error {
...
+    cx := c.(connector)
+    return d.createConnFromParams(cx.PoolParams, cx.ConnParams)
 }
 
 func (d *drv) ClientVersion() (VersionInfo, error) {
 	return d.clientVersion, nil
 }
 
-var cUTF8, cDriverName = C.CString("AL32UTF8"), C.CString(DriverName)
+var cUTF8, cDriverName = C.CString("UTF-8"), C.CString(DriverName)
```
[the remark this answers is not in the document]
anthony-tuininga: Yes, Oracle does call the standard UTF-8 as "AL32UTF8". ODPI-C, however, has to do a lookup to figure that out. If you pass "UTF-8", ODPI-C recognizes it and avoids the lookup -- so a slight performance improvement and a little easier for non-Oracle people to read and understand, I think!

And yes, "UTF8" is the non-standard "CESU-8" (note the missing dash). And yes, that is confusing, but Oracle has been around for the entire Unicode journey and is also reluctant to remove such confusing things (as they may be used by older code which would be difficult to change). So we're stuck with it!

conn.go
```diff
@@ -724,33 +719,28 @@ type TraceTag struct {
...
+// ContextWithParams()
+//   Returns a context with the specified parameters. These parameters are used
+// to modify the session acquired from the pool. If a standalone connection is
+// being used this will have no effect.
+//-----------------------------------------------------------------------------
+func ContextWithParams(ctx context.Context, params ConnParams) context.Context {
+    return context.WithValue(ctx, paramsCtxKey, params)
 }
 
-func (c *conn) ensureContextUser(ctx context.Context) error {
```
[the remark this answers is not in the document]
anthony-tuininga: Take a look at connector.Connect(). That accepts the context and the parameters are acquired from that context and used to create the connection. As such, there should be no need to "ensure" the context user as it should already be ensured. It does mean that you can't *change* the context user once you have the connection -- but I think that is appropriate. Would you agree?

anthony-tuininga:
Thanks for your comments. I'm glad you're happy with the general direction! I've pushed the changes after running "go fmt" as requested and have responded to your questions.

The only other question I have is whether we should simply modify godror.NewConnector() or leave it as godror.NewConnector2().

tgulacsi:
Just replace NewConnector with NewConnect2 - also I'll replace the innards of ConnectionParams or get rid of it althogether (its  combination of ConnParams and PoolParams).

anthony-tuininga:
Just to be clear: you want to remove the existing NewConnector() and rename NewConnector2() to NewConnector()?

tgulacsi:
More or less yes - godror.NewConnector is not part of any interface, it's a convenience function.
NewConnector2 (ATM) allows more specific setting of all connection/pool params.
The old NewConnector added only the onInit function setting, and the NewConnector2 allows this too.

If one wants to use the old connection string way, just needs a ParseConnString beforehand.

tgulacsi:
BTW I've added some changes right on top of anthony-tuininga:sharding_poc as godror:sharding_poc - less pointers and getting rid of ConnectionParams as a different struct.

anthony-tuininga:
Ok. I'll change the NewConnector2() to be NewConnector() as suggested. Did you see my other comments to your questions and are you satisfied with the responses?

conn.go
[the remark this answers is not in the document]
tgulacsi: Ok, I've had to check, but it seems that if the driver implements driver.DriverContext (has an OpenConnector method), then it will be used instead of Open.

So this shoukd work - with the trivial exception when the user has a fixed sql.Conn or sql.Tx instead of an sql.DB.

drv.go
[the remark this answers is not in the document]
tgulacsi: Thanks for the clarification, I'va added a comment about it.

tgulacsi:
Thank you for your hard work, it resolved a cleaner code structure, which I appreciate much.

The only remaining glitch is TestHeterogenousPoolIntegration in `z_heterogenous_test.go`, which also uses the dreaded _ORACLE_SCRIPT hack. 
When you have time, please help to resolve this, too!

anthony-tuininga:
You're very welcome! And yes, I do plan to work on that test script as well. :-)

Issue #37, opened 2020-03-31
Doesn't currently work with go mod vendor


veqryn:
**Describe the bug**
So go modules don't appear to contain anything directories that don't have go source files.
This means that when using `go get` to download this package, all the odpi-C files are not included, causing the package to fail at build time.

**To Reproduce**
1. Have a `go.mod` file at your project's root directory.
2. From that root directory, run this to download godror: `go get -u github.com/godror/godror`
3. Run `go mod vendor` to have godror vendored.
4. Build project, get this or similar error:
```
# github.com/godror/godror
vendor\github.com\godror\godror\conn.go:10:10: fatal error: dpiImpl.h: No such file or directory
 #include "dpiImpl.h"
          ^~~~~~~~~~~
compilation terminated.
```

**Expected behavior**
Successful build.

**Your oracle client version**
18.3.0.0.0

**Your godror version**
v0.13.2

**Your go version**
1.14.1

**Your gcc version**
7.3.0

**Machine (please complete the following information):**
 - OS: Windows
 - Architecture: x86_64
 - Version: 10

**Additional context**
There appear to be two possible solutions:

1. Include a mostly empty go file in all directories containing ODPI-C files.
Call it `required.go` or something, and just have some notes in it.
This will cause go mod to include the whole directory and all files in the directory in the module.

or 2. Include documentation on how to download and install the correct ODPI-C files onto your host or container image, so that this project will work.

veqryn:
I'd like to add that there is precedent for including .go files beside your non-go dependencies.
For example, the golang protobuf libraries include the generated golang files right beside the .proto files, that way the .proto files will end up in the module.

veqryn:
Made a PR to solve this, here: https://github.com/godror/godror/pull/38

Ran into an issue with `C source files not allowed when not using cgo`, but got around it by using build tags, yay.

tgulacsi:
Thanks!. Merged.

PR #38, opened 2020-03-31
Require ODPI-C directories to be included in vendored go modules

veqryn:
Gets the ODPI-C directories to be included in vendored golang modules (ie: `go mod vendor`).

Solves: https://github.com/godror/godror/issues/37

There is precedent to use build tags like this, see here: https://github.com/golang/go/issues/25922

I have confirmed this works in production.

Let me know if you need help or anything else to get this merged. I think a lot of people will appreciate this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/godror/godror/38)


veqryn:
Thanks. Already am on there from some previous PR's.

PR #41, opened 2020-04-01
Ability to override client timezone by adding it to connection string

andrei-liashenko:
The timezone specified in the connection string is parsed and stored to connection-params, but never actually used - timezone/offset of connection are always overriden by dbtimezone/systimestamp' s offset combination.
This pull request makes connection string's timezone to be actually usable.


andrei-liashenko:
Added setting local timezone to connection string parse.

PR #53, opened 2020-05-08
Fix memory leak

sudarshan12s:
A sample GO application with single goroutine is tried with Connection acquire and release in a continous loop.

Tried with 1 million iterations, RSS of process keeps increasing and doesnt reduce even after goroutine completes all iterations and exits.

pseudo code

i <- 1, 1000000

conn, err := db.Conn(ctx)

   if err != nil {                                                       
      log.Fatal(err)                                                      
    }                                                                     
conn.Close()

After checking the connection acquire/release code walk through, found the following variable, Connclass was not freed.

cConnClass is allocated in C heap using CString but C.Free is missed. it happens since colon (:) in line creates new variable and shadows the earlier declared variable with same name.

cConnClass := C.CString(P.ConnClass)

Tests done:

without fix:
connection acquire/release for 1 million , RSS is 74204 MB

with fix:
connection acquire/release for 1 million , RSS is 36552 MB

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,6 +17,7 @@ Ricardo Vegas <rvegas@users.noreply.github.com>
 Robin van Duiven <robin@aqua-it.com>
 Ryan Bastic <ryan@gooddoglabs.com>
 Saimon <semyuon@gmail.com>
+Sudarshan Soma <sudarshan.ss.s@oracle.com>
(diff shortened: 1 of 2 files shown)


anthony-tuininga:
@tgulacsi, just so you are aware, Sudarshan is a colleague of mine and @cjbj. This is a very simple bug fix, I think!

tgulacsi:
Thanks a lot!

PR #60, opened 2020-06-08
Add support for sharding with RAW columns.

anthony-tuininga:


drv.go
```diff
@@ -356,6 +356,13 @@ func (d *drv) acquireConn(pool *connPool, P commonAndConnParams) (*C.dpiConn, bo
 				cs := C.CString(value)
 				defer C.free(unsafe.Pointer(cs))
 				C.dpiData_setBytes(&tempData, cs, C.uint32_t(len(value)))
+			case []byte:
+				columns[i].oracleTypeNum = C.DPI_ORACLE_TYPE_RAW
+				columns[i].nativeTypeNum = C.DPI_NATIVE_TYPE_BYTES
+				bytesVal := value.([]byte)
```
tgulacsi, line 362: value is already []byte, you don't need the conversion and bytesVal.

drv.go
```diff
@@ -356,6 +356,13 @@ func (d *drv) acquireConn(pool *connPool, P commonAndConnParams) (*C.dpiConn, bo
...
+				cs := (*C.char)(C.CBytes(bytesVal))
```
tgulacsi, line 363: AFAIK tempData won't hold on to our pointer, so this copying is unneeded:
`C.dpiData_setBytes(&tempData, (*C.char)(unsafe.Pointer(&value[0])), len(value))`
is enough

anthony-tuininga: Will test that. Thanks.

drv.go
anthony-tuininga: True. Will change this.

drv.go
anthony-tuininga: Works fine. Thanks.

tgulacsi:
@anthony-tuininga , I assumed that dpiData_setBytes copies the data.
But looking into the implementation, it just stores the pointer and the length.

So I gave you a wrong advice: that sharding key should be copied, and **never freed** - this holds for the string case, too!

What is the lifetime of `tempData.value` there?

I've tried to fix this in 699ac88 .

anthony-tuininga:
I see that. Yes, the buffers in the sharding key columns (which are part of the dpiConnCreateParams structure) need to remain valid until after the call to dpiConn_create() is completed. The changes you have made make sense to me. I'll have some tests run to make sure nothing untoward was introduced accidentally.

PR #77, opened 2020-07-31
logfmt-encoded connection string

tgulacsi:
Allow appending logfmt-encoded parameters to a new line after the connection string,
or encode the DSN as "connectString".

Move all this logic to a separate "connstr" package for fuzzing (go-fuzz does not like cgo).
Keep the types as aliases and ParseConnString, NewPassword functions under godror, too.

Rename DSN to ConnectString - backwards-incompatible change!
But DSN is quite cryptic..


cjbj:
Are username & password always needed - what is needed when external authentication is used?

cjbj:
@tgulacsi here's some potential text for the README.  It's just untested suggestions!  

It would be good to change the connection string examples to show them in use in sql.Open() with the correct quoting. 

------------------

## Connect

Connect to Oracle Database using `sql.Open("godror", dataSourceName)` where
`dataSourceName` contains options such as the user credentials, the database
connection string, and other configuration settings.  It should be a
[logfmt](https://brandur.org/logfmt)-encoded parameter list.  For example:

```
user="scott" password="tiger" connectString="dbhost:1521/orclpdb1" \
poolSessionTimeout=42s configDir=/tmp/admin heterogeneousPool=false standaloneConnection=false
```

All godror parameters ([see
here](https://pkg.go.dev/github.com/godror/godror?tab=doc#pkg-overview)) should
also be specified logfmt-ted.

### Connection Strings

The `connectString` parameter can be one of:

- An Easy Connect String

  [Easy Connect
  strings](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-B0437826-43C1-49EC-A94D-B650B6A4A6EE)
  like the one shown above are most convenient to use.  The syntax allows a
  number of options to be set without requiring an external `tnsnames.ora`
  configuration file.  A more complex example usable with Oracle Client
  libraries 19c or later is:

  ```
  connectString="tcps://salesserver1:1521/sales.us.example.com?ssl_server_cert_dn=\"cn=sales,cn=Oracle Context Server,dc=us,dc=example,dc=com\"&sdu=8128&connect_timeout=60"
  ```

  Common options are `connect_timeout` to return an error if connection takes
  too long, and `expire_time` to make sure idle connections are not closed by
  firewalls.

  The technical article [Oracle Database 19c Easy Connect Plus Configurable
  Database Connection
  Syntax](https://download.oracle.com/ocomdocs/global/Oracle-Net-19c-Easy-Connect-Plus.pdf)
  contains more information.

- An Oracle Net Connect Descriptor String

  Full descriptors can be used, such as
  `connectString="(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=dbhost.example.com)(PORT=1521))(CONNECT_DATA=(SERVICE_NAME=orclpdb1)))"`

- An Oracle Net Service Name

  Connect Descriptor Strings are commonly stored in a `tnsnames.ora` file and
  associated with a Net Service Name.  This name can be used directly for
  `connectString`.  For example, given a `tnsnames.ora` file with the following
  contents:

      ORCLPDB1 =
        (DESCRIPTION =
          (ADDRESS = (PROTOCOL = TCP)(HOST = dbhost.example.com)(PORT = 1521))
          (CONNECT_DATA =
            (SERVER = DEDICATED)
            (SERVICE_NAME = orclpdb1)
          )
        )

  then you could connect using `connectString=ORCLPDB1`

  See "Optional Oracle Net Configuration Files" below.

For more connection details, see [connstr/README.md](./connstr/README.md).

You can provide all possible options with `ConnectionParams`.  Note
`ConnectionParams.String()` *redacts* the password (for security, to avoid
logging it - see <https://github.com/go-goracle/goracle/issues/79>).  If you
need the password, then use `ConnectionParams.StringWithPassword()`.

### Optional Oracle Net Configuration Files

Optional Oracle Net configuration files are used by the Oracle Client libraries
during the first call to `sql.Open`.  The directory containing the files can be
specified in the `sql.Open()` data source name with the `configDir` option.

The common files are:

* `tnsnames.ora`: A configuration file that defines databases aliases for
  establishing connections.

* `sqlnet.ora`: A profile configuration file that may contain information on
  features such as connection failover, network encryption, logging, and
  tracing.  See [Oracle Net Services
  Reference](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-19423B71-3F6C-430F-84CC-18145CC2A818)
  for more information.

The files should be in a directory accessible to godror, not on the database
server host.

If the `configDir` connection option is not used, then a search heuristic is
used to find the files.  The search includes:

* `$TNS_ADMIN`
* `/opt/oracle/instantclient_19_8/network/admin` if Instant Client is in `/opt/oracle/instantclient_19_8`.
* `/usr/lib/oracle/19.8/client64/lib/network/admin` if Oracle 19.8 Instant Client RPMs are used on Linux.
* `$ORACLE_HOME/network/admin` if godror is using libraries from a database installation.

### Oracle Session Pooling

Set `standaloneConnection=0`- this is the default.  The old advice of setting
`db.SetMaxIdleConns(0)` are obsolete with Go 1.14.6.  It does no harm, but the
revised connection pooling (synchronous ResetSession before pooling the
connection) eliminates the need for it.

To use heterogeneous pools, set `heterogeneousPool=1` and provide the username
and password through `godror.ContextWithUserPassw` or
`godror.ContextWithParams`.

tgulacsi:
> On 07/31/2020 8:43 AM Christopher Jones <notifications@github.com> wrote:
>
>
> Are username & password always needed - what is needed when external authentication is used?

Of course they can left empty or left out entirely - I just couldn't make the sentence simple enough when tried to put all corner cases in it.

tgulacsi:
Thanks @cjbj for the nice README (https://github.com/godror/godror/pull/77#issuecomment-666977121)! 
I've added it to connstr/README, as I think it's too much, too detailed for the main README.

cjbj:
> Thanks @cjbj for the nice README ([#77 (comment)](https://github.com/godror/godror/pull/77#issuecomment-666977121))!
> I've added it to connstr/README, as I think it's too much, too detailed for the main README.

Should you remove the session pooling bit from connstr/README since that is for/from the top level README?

tgulacsi:
I've hesitated over it, as it's rightfully a parameter thing, configured through the connection string.
Leaved it there, for the Go version thing and Go pool disabling warning. 
Although some rewording may help

connstring.go
```diff
@@ -0,0 +1,400 @@
...
+		if u.Port() != "" {
+			P.DSN += ":" + u.Port()
+		}
+		if u.Path != "" && u.Path != "/" {
+			P.DSN += u.Path
+		}
+		q = u.Query()
+	}
+	if paramsString != "" {
+		d := logfmt.NewDecoder(strings.NewReader(paramsString))
```
sudarshan12s: logfmt does it support lists like ShardingKey ? may be we dont need it in this format?

sudarshan12s: comment can be can be ignored, i believed lists is complex to handle as arguments. connector/openDb interface should  be able to acheive this.

tgulacsi: What kind of lists are you thinking about?
I don't know of any parameter that accepts a list, at the moment.

connstr/connstr.go
```diff
@@ -0,0 +1,683 @@
...
+	var paramsString string
+	dataSourceName = strings.TrimSpace(dataSourceName)
+	if i := strings.IndexByte(dataSourceName, '\n'); i >= 0 {
+		dataSourceName, paramsString = strings.TrimSpace(dataSourceName[:i]), strings.TrimSpace(dataSourceName[i+1:])
+	}
+	var q url.Values
+
+	if !strings.HasPrefix(dataSourceName, "oracle://") {
+		// Not URL
+		q = make(url.Values, 32)
```
anthony-tuininga, line 290: The 32 here is used above as well and should probably be a constant of some kind?

connstr/corpus/000cfee36900daff3b50289be68e9cfdd1b6d416-15
```diff
@@ -0,0 +1 @@
+oracle://%ae+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
anthony-tuininga, line 1: What are all of these corpus files?

drv.go
```diff
@@ -5,30 +5,30 @@
...
-//     connectionClass=& \
-//     standaloneConnection=0& \
-//     enableEvents=0& \
-//     heterogeneousPool=0& \
-//     prelim=0& \
-//     poolWaitTimeout=5m& \
-//     poolSessionMaxLifetime=1h& \
-//     poolSessionTimeout=30s& \
-//     timezone=Local& \
+//   sid/? \n
```
anthony-tuininga, line 14: Did this get corrupted?

drv.go
```diff
@@ -5,30 +5,30 @@
 
 // Package godror is a database/sql/driver for Oracle DB.
 //
-// The connection string for the sql.Open("godror", connString) call can be
+// The connection string for the sql.Open("godror", connstr) call can be
```
anthony-tuininga, line 8: This should say "dataSourceName" instead of connstr. Below instead of sid it should say connectString.

connstring.go
anthony-tuininga: He is referring to sharding keys. You can pass multiple values although it is more common to use just one.

tgulacsi: Ah, ok. Now it accepts only one, but it's a very simple fix (see handling of onInit).

drv.go
tgulacsi: Hmm, the last version looks the same.

connstr/corpus/000cfee36900daff3b50289be68e9cfdd1b6d416-15
tgulacsi: Fuzzing corpus. I'll remove them.

connstr/connstr.go
```diff
@@ -0,0 +1,698 @@
...
+				case "connectString":
+					var user, passw string
+					if user, passw, P.ConnectString = parseUserPassw(value); P.Username == "" && P.Password.IsZero() {
+						P.Username, P.Password.secret = user, passw
+					}
+				case "user":
+					P.Username = value
+				case "password":
+					P.Password.secret = value
+				case "onInit", "shardingKey", "superShardingKey":
```
anthony-tuininga, line 353: Sharding keys can be not just strings. They can be numbers, dates, and RAW (binary) values. Does this handle only strings? Does it handle multiple values?

drv.go
```diff
@@ -5,30 +5,30 @@
...
-//     connectionClass=& \
-//     standaloneConnection=0& \
-//     enableEvents=0& \
-//     heterogeneousPool=0& \
-//     prelim=0& \
-//     poolWaitTimeout=5m& \
-//     poolSessionMaxLifetime=1h& \
-//     poolSessionTimeout=30s& \
-//     timezone=Local& \
+//   host:port/service_name? \n
```
anthony-tuininga, line 14: Isn't this supposed to be "connectString=host:port/service_name" and the trailing "?" removed? Perhaps with a footnote as you said earlier regarding the old format that is still supported?

connstr/connstr.go
tgulacsi: Yes, only strings - before this, you could specify sharding keys only directly.
How should one specify the data type? 
JSON would handle strings and numbers, but nothing else.

drv.go
tgulacsi: You're right, this is the old oracle://... format.

connstr/README.md
```diff
@@ -0,0 +1,118 @@
...
+   features such as connection failover, network encryption, logging, and
+   tracing.   See [Oracle Net Services
+   Reference](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-19423B71-3F6C-430F-84CC-18145CC2A818)
+   for more information.
+
+The files should be in a directory accessible to godror, not on the database
+server host.
+
+If the `configDir` connection option is not used, then a search heuristic is
+used to find the files.   The search includes:
```
cjbj, line 87: I would now tweak this to read "used to find the directory containing the files."

connstr/README.md
```diff
@@ -0,0 +1,118 @@
+# Connect
+
+Connect to Oracle Database using `sql.Open("godror", dataSourceName)` where
+`dataSourceName` contains options such as the user credentials, the database
+connection string, and other configuration settings.   It should be a
+[logfmt](https://brandur.org/logfmt)-encoded  parameter list.   For example:
+
```
cjbj, line 7: Now I've testing it, what about adding a 'real' example to make it easier for newbies to cut & paste? Perhaps change the example to be something like this?:

----------
For example:
```
db, err := sql.Open("godror", `user="scott" password="tiger" connectString="dbhost:1521/orclpdb1"`)
```
Other connection and driver options can also be used:

```
db, err := sql.Open("godror", `user="scott" password="tiger" connectString="dbhost:1521/orclpdb1?connect_timeout=2" poolSessionTimeout=42s configDir="/opt/oracle/configdir" heterogeneousPool=false standaloneConnection=false`)
```

---------

@tgulacsi since I don't use your older syntaxes, I find not allowing line breaks in this syntax is a bit painful.

connstr/README.md
```diff
@@ -0,0 +1,128 @@
...
+db, err := sql.Open("godror", `user="scott" password="tiger" connectString="dbhost:1521/orclpdb1" 
+	poolSessionTimeout=42s configDir=/tmp/admin 
+	heterogeneousPool=false standaloneConnection=false`)
+```
+Other connection and driver options can also be used:
+```
+db, err := sql.Open("godror", `user="scott" password="tiger" 
+	connectString="dbhost:1521/orclpdb1?connect_timeout=2" 
+	poolSessionTimeout=42s configDir="/opt/oracle/configdir" 
+	heterogeneousPool=false standaloneConnection=false`)
```
cjbj, line 19: There's a stray Ctrl-M at the end of this line.

connstr/connstr_test.go
```diff
@@ -0,0 +1,268 @@
...
+	P.Username, P.Password = "scott", NewPassword("tiger")
+	P.ConnectString = "dbhost:1521/orclpdb1?connect_timeout=2"
+	P.SessionTimeout = 42 * time.Second
+	P.SetSessionParamOnInit("NLS_NUMERIC_CHARACTERS", ",.")
+	P.SetSessionParamOnInit("NLS_LANGUAGE", "FRENCH")
+	fmt.Println(P.StringWithPassword())
+	// Output:
+	// user=scott password=tiger connectString="dbhost:1521/orclpdb1?connect_timeout=2"
+	// configDir= connectionClass= enableEvents=0 heterogeneousPool=0 libDir= newPassword=
+	// onInit="ALTER SESSION SET NLS_NUMERIC_CHARACTERS = q'(,.)'" onInit="ALTER SESSION SET NLS_LANGUAGE = q'(FRENCH)'"
```
cjbj, line 264: For performance, this would ideally execute just one statement with two options like `ALTER SESSION SET NLS_NUMERIC_CHARACTERS=q'(,.)' NLS_LANGUAGE=q'(FRENCH)'`

connstr/README.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+All godror parameters ([see
+here](https://pkg.go.dev/github.com/godror/godror?tab=doc#pkg-overview)) should
+also be specified logfmt-ted.
+
+You can use `ConnectionParams` to properly build such a string:
+
+    var P connstr.ConnectionParams
+	P.Username, P.Password = "scott", connstr.NewPassword("tiger")
```
cjbj, line 29: Maybe use `godror` here since that works in simple cases? 
```
var P godror.ConnectionParams
P.Username, P.Password = "scott", godror.NewPassword("tiger")
```

connstr/README.md
```diff
@@ -0,0 +1,145 @@
...
+	P.ConnectString = "dbhost:1521/orclpdb1?connect_timeout=2"
+	P.PoolSessionTimeout = 42 * time.Second
```
cjbj, line 31: This doesn't parse.  Should it be `P.SessionTimeout`?

Is there a full list of options that can be added to the doc here?

connstr/connstr_test.go
tgulacsi: Nice! I didn't know it.

connstr/README.md
```diff
@@ -0,0 +1,145 @@
...
+    var P godror.ConnectionParams
+	P.Username, P.Password = "scott", godror.NewPassword("tiger")
+	P.ConnectString = "dbhost:1521/orclpdb1?connect_timeout=2"
+	P.SessionTimeout = 42 * time.Second
+	P.SetSessionParamOnInit("NLS_NUMERIC_CHARACTERS", ",.")
+	P.SetSessionParamOnInit("NLS_LANGUAGE", "FRENCH")
+	fmt.Println(P.StringWithPassword())
+	db := sql.OpenDB(godror.NewConnector(P))
+
+Or if you really want to build it "by hand", user connstr.AppendLogfmt:
```
cjbj, line 37: Small typo: "user" -> "use".

Also, would it be better to rename `connstr` to `dsn`, if it is going to be passed to `Open()` as the dataSourceName parameter?

tgulacsi: Good catch! I'll rename it.

cjbj:
@tgulacsi where are you heading with the OnInit, alterSession etc design?   Without debugging, I couldn't quickly see if onInit was like cx_Oracle's [`sessionCallback`](https://cx-oracle.readthedocs.io/en/latest/user_guide/connection_handling.html#session-callbacks-for-setting-pooled-connection-state) (see [implementation](https://github.com/oracle/python-cx_Oracle/blob/master/src/cxoConnection.c#L626-L636) and [here](https://github.com/oracle/python-cx_Oracle/blob/master/src/cxoSessionPool.c#L143-L144)).  @anthony-tuininga probably has more comments, since he knows your code base better than me.

(PS does alterSession need to be mentioned in drv.go header comments?)

tgulacsi:
It should be callback-based - now it is not, uses connCreateParams.outNewSession on create/acquire. But as #78 indicates, it does not work in every case.

kurt-google:
Its somewhat unfortunate this was so readily merged.

The old format could have been checked by checking for an `oracle://` schema (Correct me if this a valid logfmt key, it seems lib/pq does just this https://godoc.org/github.com/lib/pq) prefix, and it would have been nice to have a transition period instead of requiring users to update all references at once. Earlier the only other option for specifying parameters also went through a backwards incompatible change (Though much simpler to resolve). These make it harder to keep up with all of the excellent work going into this driver.

tgulacsi:
@kurt-google the old formats are still accepted, though not advertised.

They are documented at the bottom of ./dan/README.md

kurt-google:
Thanks for clearing that up. It seems I didnt see it's new location in the diff. Sorry for the noise.

PR #78, opened 2020-08-03
WIP: Fix setting of PreFetchCount and FetchArraySize

sudarshan12s:
# Following changes are done. Currently review is in progress.

- Moved the PrefetchCount and ArrayFetchCount before statement executes. Previously, it was set after statement execute in openRows.
- Default values for prefetch, arrayfetch are handled correctly. Previously If user sets prefetch count as 0, driver replaces it with defaultvalue, DefaultPrefetchCount. same thing for arraysize.
- New behaviour
If user doesn't set prefetch or arrayfetch, they are treated as defaults, DefaultPrefetchCount and DefaultFetchArraySize respectively
If user sets 0 or negative values for prefetch, arrayfetch, they are sent as 0. For ArrayFetch, 0/-1 is invalid and ODPI error is thrown. >0 values are of course accepted and passed to the driver.

- minor correction for removing POOLED from connectionclass , left out from earlier change

- Updated README for Tuning doc

- added test case , TestPreFetchQuery in z_test.go to verify the prefetch count and array sizes.

    for running this test, we need to set GODROR_TEST_SYSTEM_USERNAME and GODROR_TEST_SYSTEM_PASSWORD.
    Ex:
    export GODROR_TEST_SYSTEM_USERNAME=system
    export GODROR_TEST_SYSTEM_PASSWORD=oracle
# Testing
- Program:
It executes single row fetch, where in exactly one row is fetched. It does multirow fetch, where 99 rows are fetched. round trips are calculated with different combinations of these alues.

case 1: user doesn't set prefetchcount and fetcharraysize.

Singlerowfetch -> 1

MultiRowfetch -> 2

case 2: user set prefetchcount=0 and fetcharraysize=100.

Singlerowfetch -> 2

MultiRowfetch -> 2

case 3: user set prefetchcount=1 and fetcharraysize=100.

Singlerowfetch -> 2

MultiRowfetch -> 2

case 4: user set prefetchcount=2 and fetcharraysize=100.

Singlerowfetch -> 1

MultiRowfetch -> 2

case 5: user set prefetchcount=100 and fetcharraysize=100.

Singlerowfetch -> 1

MultiRowfetch -> 1

go test -v -run TestPreFetchQuery
=== RUN   TestPreFetchQuery
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0

    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: -1 , Arraysize: -1 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: -1 , Arraysize: -1 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 0 , Arraysize: -1 , Rows:  1 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 0 , Arraysize: -1 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: -1 , Arraysize: 100 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: -1 , Arraysize: 100 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 0 , Arraysize: 100 , Rows:  1 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 0 , Arraysize: 100 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 1 , Arraysize: 100 , Rows:  1 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 1 , Arraysize: 100 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 2 , Arraysize: 100 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 2 , Arraysize: 100 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 100 , Arraysize: 100 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 100 , Arraysize: 100 , Rows:  99 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 1 , Arraysize: -1 , Rows:  1 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 1 , Arraysize: -1 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 2 , Arraysize: -1 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 2 , Arraysize: -1 , Rows:  99 , Round-trips: 2
    TestPreFetchQuery: z_test.go:3276: godror : v0.18.0
    TestPreFetchQuery: z_test.go:3343: SingleRowFetch:  Prefetch: 100 , Arraysize: -1 , Rows:  1 , Round-trips: 1
    TestPreFetchQuery: z_test.go:3349: MultiRowFetch:  Prefetch: 100 , Arraysize: -1 , Rows:  99 , Round-trips: 1
--- PASS: TestPreFetchQuery (0.43s)
PASS
ok  	github.com/godror/godror	0.921s

- With -1/0 ArraySize, ODPI error is thrown
2020/07/31 09:00:58 Next: ORA-00000: DPI-1018: array size of 1 is too small.


stmt.go
```diff
@@ -150,18 +160,24 @@ func FetchRowCount(rowCount int) Option { return FetchArraySize(rowCount) }
 
 // FetchArraySize returns an option to set the rows to be fetched, overriding DefaultFetchRowCount.
 func FetchArraySize(rowCount int) Option {
```
cjbj, line 162: For node-oracledb, we throw a node-oracledb error when the arraysize is 0.  This is a little nicer than a DPI error because it can give the specific parameter name.  Is that something to think about for godror?

z_test.go
```diff
@@ -3185,3 +3193,227 @@ func TestOpenCloseLob(t *testing.T) {
...
+		var sid uint
+		sql := "SELECT sys_context('userenv','sid') FROM dual"
+		err := testDb.QueryRow(sql).Scan(&sid)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Get godror version number
+		var driver string
+		sql2 := `SELECT UNIQUE client_driver
```
cjbj, line 3265: Getting the godror version number isn't needed; I has used it for ad-hoc testing.

z_test.go
```diff
@@ -3185,3 +3193,227 @@ func TestOpenCloseLob(t *testing.T) {
...
+	}
+
+	ctx, cancel := context.WithTimeout(testContext("TestPreFetchQuery"), 30*time.Second)
+	defer cancel()
+
+	// Create a table used for Prefetch, ArrayFetch queries
+
+	tbl := "t_employees" + tblSuffix
+	testDb.ExecContext(ctx, "DROP TABLE "+tbl)
+	if _, err := testDb.ExecContext(ctx, "CREATE TABLE "+tbl+" (employee_id NUMBER, last_name VARCHAR2(30))"); err != nil {
```
cjbj, line 3215: The table doesn't need two columns.  The test would be simpler with just a single column.

z_test.go
```diff
@@ -3185,3 +3193,227 @@ func TestOpenCloseLob(t *testing.T) {
...
+		t.Log(driver)
+
+		return sid
+
+	}
+
+	// verify round trips for SingleRowFetch and MultiRowFetch
+	// and return failure on unexpected roundtrips
+
+	srt, mrt := runPreFetchTests(t, sid(), -1, -1) // prefetch, arraysize not given
```
cjbj, line 3281: Thinking more about this, there will need to be a way to set the parameters to negative numbers (instead of using -1 to indicate 'use the default'). This is needed so error handling of invalid values can be tested.

stmt.go
```diff
@@ -40,8 +40,8 @@ import (
 
 type stmtOptions struct {
 	boolString         boolString
-	fetchArraySize     int
-	prefetchCount      int
+	fetchArraySize     int // zero means DefaultFetchArraySize negative means 0
+	prefetchCount      int // zero means DefaultPrefetchCount
```
cjbj, line 44: And what does negative mean?  

Overall the definitions do seem a little complex, as @anthony-tuininga noted.  I know you said you were trying to be consistent with other values.  Maybe @tgulacsi has some comments.

sudarshan12s: The behaviour of these attributes is kept similar to the way we handle MaxIdleConnections.

z_test.go
sudarshan12s: thought if someone wants to print the client driver version, code snipped would help. should we add seperate testcase?

z_test.go
```diff
@@ -3185,3 +3193,227 @@ func TestOpenCloseLob(t *testing.T) {
...
+
+	var r uint
+	// Do some work
+	r = singleRowFetch(t, pf, as)
+
+	rt2 := getRoundTrips(t, sid)
+
+	t.Log("SingleRowFetch: ", "Prefetch:", pf, ", Arraysize:", as, ", Rows: ", r, ", Round-trips:", rt2-rt1)
+	srt := rt2 - rt1
+	rt1 = getRoundTrips(t, sid)
```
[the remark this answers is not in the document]
sudarshan12s: Yes incorporated ...

README.md
```diff
@@ -65,25 +65,147 @@ with OUT parameters, or sending/retrieving PL/SQL array types - just give a
 
 The array size of the returned PL/SQL arrays can be set with `godror.ArraySize(2000)` (default value is 1024).
 
-## Speed
+## Tuning
```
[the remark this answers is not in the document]
tgulacsi: Maybe. What should be grouped together in another documentation? Options?

stmt.go
tgulacsi: I think zero array size won't work, so we can leave 0 as default, any other as is.

stmt.go
tgulacsi: Zero array size is uninterpretable, so leave 0 as default, negative is an error (or default).

README.md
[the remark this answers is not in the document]
cjbj: @tgulacsi How is https://github.com/golang/go/tree/master/doc generated- by hand?  I used a single page doc format https://github.com/oracle/node-oracledb/blob/master/doc/api.md for Node.js.  As a reader I like this since I can do quick browser Ctrl-F 'find' page searches.  As a maintainer it means doing a lot of renumbering since, with all the nesting in Node.js, there are a lot of subheadings and numbered sections help a bit to show the indentation level of the heading. Or you could add some doc generation tooling and put doc in multiple files, like https://github.com/oracle/python-cx_Oracle/tree/master/doc.

z_test.go
cjbj: That's out of scope for this PR :)

z_test.go
cjbj: Maybe make something like -99 mean use the default.  This lets -1 be used in boundary tests to make sure error handling is correct.

README.md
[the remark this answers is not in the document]
tgulacsi: "Tuning" is a good self-contained part, gives itself to separate it.

stmt.go
tgulacsi: fetchArraySize cannot be 0.
prefetchCount can be 0, that was the default before ODPI v4.

z_test.go
```diff
@@ -3185,3 +3193,227 @@ func TestOpenCloseLob(t *testing.T) {
...
+
+	srt, mrt = runPreFetchTests(t, sid(), 2, -1) // array size not given
+	if srt != 1 || mrt != 2 {
+		t.Fatal("wanted 1 SingleFetchRoundTrip and 2 MultiFetchRoundTrip, got ", srt, " SingleFetchRoundTrip and ", mrt, " MultiFetchRoundTrip")
+	}
+
+	srt, mrt = runPreFetchTests(t, sid(), 100, -1) // array size not given
+	if srt != 1 || mrt != 1 {
+		t.Fatal("wanted 1 SingleFetchRoundTrip and 1 MultiFetchRoundTrip, got ", srt, " SingleFetchRoundTrip and ", mrt, " MultiFetchRoundTrip")
+	}
```
tgulacsi, line 3327: Use a table test:

```
for _, tCase := range []struct { pf, as, srt, mrt int }{
    {-1, -1,  1, 2},
    {0, -1, 2, 2},
    ...
}{
  srt, mrt := runPreFetchTest(t, sid(), tCase.pf, tCase.as)
  if !(srt == tCase.srt && mrt = tCase.mrt) {
    t.Fatalf("wanted %d/%d SingleFetchRoundTrip / MultiFetchRoundTrip, got %d/%d", tCase.srt, tCase.mrt, srt, mrt)
  }
}
```

README.md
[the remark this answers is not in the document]
cjbj: @tgulacsi Let me know what you want to do with it (how you want it separated).  See my other note about doc.

stmt.go
cjbj: ODPI would actually have used prefetch of 2 since that was hardcoded. The ODPI v4 default prefetch is still 2.  But now in ODPI 4 the prefetch can be set by the user to >= 0.  How do you want this represented in godror?

README.md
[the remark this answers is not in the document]
cjbj: Let us know how you want it separated.

stmt.go
sudarshan12s: I will try to clarify the comments with few details . Please see

Go initializes all variables inside struct to 0. so if user doesnt overwrite these parameters, fetchArraySize and prefetchCount will be zero. in the comment, when 0 is mentioned, it points to Go initialization. By default they have 0, but when we sent to OCI, they are sent as DefaultPrefetchCount and DefaultArraySize values respectively. 

Now if user overwrites it to 0, then internally we modify it to -1 to differentiate from Go initialization. So when comment says -1, it  indicates that we are going to pass 0 to OCI. when user passes 0, it is passed as 0 to OCI . In case of ArraySize, it returns error. Now if user sets -1 or any negative value, we set internally as -1 and pass 0 to OCI. 

I went through code for returning error, when user sets -1 , it looks not trivial change. @tgulacsi  , can you correct?

z_test.go
sudarshan12s: Yes removed.

z_test.go
sudarshan12s: modified to use single column

z_test.go
sudarshan12s: Thanks. Changed to the way suggested .

stmt.go
tgulacsi: If the user does not set anything, then stmtOptions.fetchArraySize will be 0. 
That should mean DefaultArraySize.

For prefetchCount, 0 means uninitialized, which should also mean the default, DefaultPrefetchCount.
In the PrefetchCount option, we can store -1, which can mean 0.

What if we store prefetchCount+1, and subtract 1 in stmtOption.PrefetchCount() ?
then 0 is the uninitialized -> DefaultPrefetchCount,
and anything else is one more than what has been set through the PrefetchCount option.

sudarshan12s: If i understand correctly, above will allow DefaultPrefetchCount to be sent if user hasn't set anything. If he sets 0, it passes 0 to OCI. If he sets -1/-2, do we plan to throw error or treat is as 0 or defaults like DefaultPrefetchCount .

should we throw error for -1/-2,.. values from go driver?

current patch achieves the above requirement, it treats -1/-2 as 0. no error is thrown. however  error is thrown for 0 ArraySize from ODPI. For PrefetchCount, 0 is valid.

tgulacsi: The truth is that we can use ANYTHING in stmtOptions - as it is only set through the godror.PretechCount option (function), that can encode the user supplied value  in any way.
This encoding is what stmtOption.PrefetchCount should interpret and return the ready-to-use value.

For user setting, negative values are uninterpretable, so use the default.
Zero should mean zero.
Then we can encode it as "-1 means zero", "zero means default", "positive value means itself".
Or as I suggested, "zero means default", "positive means one more than the real value".
Both are acceptable, I favor mine :)

I just want to keep the 0 mean "not set by the user, so use the default", to let the uninitialized stmtOption be usable as is.

TL;DR; 
  - godror.PrefetchCount option should accept anything, <=0 means Default, >0 means  that number;
  - encode it as you feel in stmtOption.prefetchCount, but 0 must mean default;
  - stmtOption.PrefetchCount must decode this and return the number to be passed on to ODPI.

sudarshan12s: when you say, <=0 means Default,>0 means that number. how does user pass 0. 
0 is special right? if user sets 0, he meant to disable prefetch which cant be DefaultPrefetchCount. if he doesnt set anything, we take default, DefaultPrefetchCount.

1. applying your logic, user sets negative, it goes to default. if he sets 0, it gets disabled. else number passed as is
user doesnt set anything, we get -1 (0-1) in stmtOption.prefetchCount. we pass DefaultPrefetchCount down
user sets -1, we get -1 (0-1) in stmtOption.prefetchCount. we pass DefaultPrefetchCount down
user sets 0, we get 0 in stmtOption.prefetchCount. we pass 0 down

2. I was doing this:
if user sets 0 or -1, disable PreFetchCount. else number. passed as is


you see option 2 not good :)

tgulacsi: You seem to miss the fact that stmtOption.prefetchCount is an encoded value, which is set by godror.PrefetchCount, read by stmtOption.PrefetchCount - both controlled by us.
So there's no "user sets ...", only "user gives n to godror.PrefetchCount, which stores m, which is decoded to k by stmtOption.PrefetchCount, which is used as-is".

PR #129, opened 2021-01-29
modify ContextWithUserPassw to retain values set in pool creation

sudarshan12s:
ContextWithUserPassw was setting the godror parameters such as Timezone, EnableEvents to default values. 
so if we do QueryRowContext with ctx derived from godror.ContextWithUserPassw , it would change the godror CommonParams, ConnParams which results in creating new pool as the poolKey changes.

Fixes are made to retain CommonParams, ConnParams of the connection and overwrite the username, password, connectionclass parameters passed .

Tests:

go test  -run TestContextWithUserPassw
go test  -run TestHeterogeneousPoolIntegration

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -855,8 +855,16 @@ func (tt TraceTag) String() string {
 	return q.Encode()
 }
 
-const paramsCtxKey = ctxKey("params")
+const paramsCtxKey              = ctxKey("params")
+const userPasswdConnClassCtxKey = ctxKey("userPasswdConnClass")
 
+// UserPasswdConnClassTag consists of Username, Password 
+// and ConnectionClass values that can be set with ContextWithUserPassw
+type UserPasswdConnClassTag struct {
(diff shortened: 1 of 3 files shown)


tgulacsi:
Sorry, I've also submitted a solution, but misses  UserPasswdConnClassTag.
Please start with f6efcc1 and add UserPasswdConnClassTag in ContextWithUserPassw, and simpolify the uses where userPasswCtxKey is read.

sudarshan12s:
> Sorry, I've also submitted a solution, but misses UserPasswdConnClassTag.
> Please start with [f6efcc1](https://github.com/godror/godror/commit/f6efcc1969c44ee1239e389ecc5d7acefb181c6e) and add UserPasswdConnClassTag in ContextWithUserPassw, and simpolify the uses where userPasswCtxKey is read.

No issues, will incorporate comment.

PR #196, opened 2021-11-08
33159821 rev2

AyushMall124:
TestStorageTypes tests and verifies the JSON-type each supported Go-type is stored as in the DB, when when inserted into a JSON column.
A map with all these types are converted into godor.JSONValue and bound to an insert function which is stored as a JSON-object in the DB. From the server the JSON-type of each field of this object is fetched and compared with the expected type.

<head>

</head>

<body>

Go | Json Type
-- | --
Float32, float64 | Number
Uint64 | Number
Int, int8, int16, int32, int64 | Number
Time.time | timestamp
[]byte | Binary
string | String
godror.Number | number
Bool | boolean

</body>

</html>

### DIFF
--- a/z_json_test.go
+++ b/z_json_test.go
@@ -774,12 +774,27 @@ func TestJSONStorageTypes(t *testing.T) {
 		t.Errorf("%d/1. (%v): %v", 0, jsmap, err)
 
 	}
+	keys := []string{ "asNumber",
+                  "asString",
+                  "asTimestamp",
+                  "asBoolean",
+                  "asByte",
+                  "asInt32",
+                  "asInt64",
+                  "asInt8",
+                  "asInt16",
+                  "asUint64",
+                  "asFloat64",
+                  "asFloat32"}
 
-	rows, err := conn.QueryContext(ctx,
-		"SELECT id, json_value(jdoc,'$.asObject.type()'), json_value(jdoc,'$.asObject.asString.type()'), json_value(jdoc,'$.asObject.asNumber.type()'), json_value(jdoc,'$.asObject.asBoolean.type()'), json_value(jdoc,'$.asObject.asTimestamp.type()'), json_value(jdoc,'$.asObject.asInt32.type()'), json_value(jdoc,'$.asObject.asInt64.type()'), json_value(jdoc,'$.asObject.asInt8.type()'), json_value(jdoc,'$.asObject.asInt16.type()'), json_value(jdoc,'$.asObject.asUint64.type()'), json_value(jdoc,'$.asObject.asFloat64.type()'), json_value(jdoc,'$.asObject.asFloat32.type()'), json_value(jdoc,'$.asObject.asByte.type()') FROM "+tbl+" c ") //nolint:gas
+	stmt_str := "SELECT id, json_value(jdoc,'$.asObject.type()')"
+	for _, key := range keys {
+		stmt_str +=  fmt.Sprintf(", json_value(jdoc,'$.asObject.%s.type()')", key)
+	}
+	stmt_str += " FROM "+tbl+ " c "
+	rows, err := conn.QueryContext(ctx,stmt_str)
 	if err != nil {
 		t.Errorf("%d/3. %v", 0, err)
-
 	}
 	defer rows.Close()
 	for rows.Next() {
@@ -789,10 +804,10 @@ func TestJSONStorageTypes(t *testing.T) {
 			asFloat64type, asFloat32type, asBytetype string
 
 		err = rows.Scan(&id, &asObjecttype,
-			&asStringtype, &asNumbertype, &asBooleantype,
-			&asTimestamptype, &asInt32type, &asInt64type,
-			&asInt8type, &asInt16type, &asUint64type,
-			&asFloat64type, &asFloat32type, &asBytetype)
+			&asNumbertype, &asStringtype, &asTimestamptype,
+			&asBooleantype, &asBytetype, &asInt32type,
+			&asInt64type, &asInt8type, &asInt16type,
+			&asUint64type, &asFloat64type, &asFloat32type)
 		if err != nil {
 			t.Errorf("%d/3. scan: %v", 0, err)
 			break
--- a/z_json_test.go
+++ b/z_json_test.go
@@ -797,73 +797,32 @@ func TestStorageTypes(t *testing.T) {
 		wantBooleanDBtype := "boolean"
 		wantBinaryDBtype := "binary"
 
-		// Validate Go-type with corresponding DB type
-		if asObjecttype != wantObjectDBtype {
-			t.Errorf("Got %+v, wanted %+v", asObjecttype,
-				wantObjectDBtype)
-		}
+                for _, tCase := range []struct {
+                       goType string
+                       getType , wantType string
+                }{
+                        {"map" , asObjecttype, wantObjectDBtype},
+                        {"string", asStringtype, wantStringDBtype},
+                        {"godror.number", asNumbertype, wantNumberDBtype},
+                        {"boolean", asBooleantype, wantBooleanDBtype},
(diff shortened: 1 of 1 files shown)


z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
 	}
 }
 
+// Converts map with different Go-types to godror.JSON type
+// This is bound to the insert function and executed
+// For each unique go-type in the map, their corresponding JSON types,
+// as stored in the DB, are fetched and compared with their expected values.
+func TestStorageTypes(t *testing.T) {
```
tgulacsi, line 712: Use a more specific name: TestJSONStorageTypes as this test needs a high enough DB version.

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer conn.Close()
+	tbl := "test_objectcollection_jsonmap" + tblSuffix
+	conn.ExecContext(ctx, "DROP TABLE "+tbl)
+	_, err = conn.ExecContext(ctx,
+		"CREATE TABLE "+tbl+" (id NUMBER(6), jdoc JSON)", //nolint:gas
+	)
+	if err != nil {
+		t.Fatal(err)
```
tgulacsi, line 728: t.Skip if the error is that the DB does not have JSON column type (old)!
(See line 48)

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+	}
+
+	jsonval := godror.JSONValue{Value: jsmap}
+	if _, err = stmt.ExecContext(ctx, 0, jsonval); err != nil {
+		t.Errorf("%d/1. (%v): %v", 0, jsmap, err)
+
+	}
+
+	rows, err := conn.QueryContext(ctx,
+		"SELECT id,json_value(jdoc,'$.asObject.type()'),json_value(jdoc,'$.asObject.asString.type()'),json_value(jdoc,'$.asObject.asNumber.type()'),json_value(jdoc,'$.asObject.asBoolean.type()'),json_value(jdoc,'$.asObject.asTimestamp.type()'),json_value(jdoc,'$.asObject.asInt32.type()'),json_value(jdoc,'$.asObject.asInt64.type()'),json_value(jdoc,'$.asObject.asInt8.type()'),json_value(jdoc,'$.asObject.asInt16.type()'),json_value(jdoc,'$.asObject.asUint64.type()'),json_value(jdoc,'$.asObject.asFloat64.type()'),json_value(jdoc,'$.asObject.asFloat32.type()'),json_value(jdoc,'$.asObject.asByte.type()') FROM "+tbl+" c ") //nolint:gas
```
tgulacsi, line 771: split with spaces, not just commas.
Generate if possible (from the `jsmap["asObject"]` keys?)

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+		var id int
+		var asStringtype, asTimestamptype, asObjecttype, asBooleantype, asNumbertype, asInt32type, asInt64type,
+			asInt8type, asInt16type, asUint64type,
+			asFloat64type, asFloat32type, asBytetype string
+
+		if err = rows.Scan(&id, &asObjecttype,
+			&asStringtype, &asNumbertype, &asBooleantype,
+			&asTimestamptype, &asInt32type, &asInt64type,
+			&asInt8type, &asInt16type, &asUint64type,
+			&asFloat64type, &asFloat32type, &asBytetype); err != nil {
```
tgulacsi, line 786: Put the error check into the next line.

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+
+	}
+
+	rows, err := conn.QueryContext(ctx,
+		"SELECT id,json_value(jdoc,'$.asObject.type()'),json_value(jdoc,'$.asObject.asString.type()'),json_value(jdoc,'$.asObject.asNumber.type()'),json_value(jdoc,'$.asObject.asBoolean.type()'),json_value(jdoc,'$.asObject.asTimestamp.type()'),json_value(jdoc,'$.asObject.asInt32.type()'),json_value(jdoc,'$.asObject.asInt64.type()'),json_value(jdoc,'$.asObject.asInt8.type()'),json_value(jdoc,'$.asObject.asInt16.type()'),json_value(jdoc,'$.asObject.asUint64.type()'),json_value(jdoc,'$.asObject.asFloat64.type()'),json_value(jdoc,'$.asObject.asFloat32.type()'),json_value(jdoc,'$.asObject.asByte.type()') FROM "+tbl+" c ") //nolint:gas
+	if err != nil {
+		t.Errorf("%d/3. %v", 0, err)
+
+	}
+	for rows.Next() {
```
tgulacsi, line 776: defer rows.Close()

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+			asFloat64type, asFloat32type, asBytetype string
+
+		if err = rows.Scan(&id, &asObjecttype,
+			&asStringtype, &asNumbertype, &asBooleantype,
+			&asTimestamptype, &asInt32type, &asInt64type,
+			&asInt8type, &asInt16type, &asUint64type,
+			&asFloat64type, &asFloat32type, &asBytetype); err != nil {
+			rows.Close()
+			t.Errorf("%d/3. scan: %v", 0, err)
+			continue
```
tgulacsi, line 789: What should continue? rows is closed.
A `break` would be better, IMHO.

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+                        {"int32", asInt32type, wantNumberDBtype},
+                        {"int64", asInt64type, wantNumberDBtype},
+                        {"int8", asInt8type, wantNumberDBtype},
+                        {"int16", asInt16type, wantNumberDBtype},
+                        {"uint64", asUint64type, wantNumberDBtype},
+                        {"float64", asFloat64type, wantNumberDBtype},
+                        {"float32", asFloat32type, wantNumberDBtype},
+                        {"[]byte", asBytetype, wantBinaryDBtype},
+               } {
+                        if !(tCase.getType == tCase.wantType) {
```
tgulacsi, line 818: `if tCase.getType != tCase.wantType {`

z_json_test.go
```diff
@@ -705,6 +705,126 @@ func TestUpdateJSONScalar(t *testing.T) {
...
+
+	stmt, err := conn.PrepareContext(ctx,
+		"INSERT INTO "+tbl+" (id, jdoc) VALUES (:1, :2)", //nolint:gas
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stmt.Close()
+	//        birthdate, err := time.Parse(time.UnixDate, "Wed Feb 25 11:06:39 PST 1990")
+	birthdate := time.Now()
```
tgulacsi, line 744: Maybe a constant date with controlled time zone info would test more corner cases.
You can use the global `birthdate` variable.

tgulacsi (changes requested):

z_json_test.go
```diff
@@ -766,27 +774,45 @@ func TestStorageTypes(t *testing.T) {
...
+                  "asByte",
+                  "asInt32",
+                  "asInt64",
+                  "asInt8",
+                  "asInt16",
+                  "asUint64",
+                  "asFloat64",
+                  "asFloat32"}
+
+	stmt_str := "SELECT id, json_value(jdoc,'$.asObject.type()')"
```
tgulacsi, line 792: Use a strings.Builder.

tgulacsi (approved):

PR #203, opened 2021-12-21
Reduce allocations

ld-dev123:


tgulacsi:
Wrong solution, but this can be reduced - see

ld-dev123:
yes, that's why i closed the PR (opened by mistake) - was just a quick test for me to see if the allocations go down...Thank You !

PR #206, opened 2021-12-29
Creating godror.Object and godror.ObjectCollection from map and array of map.

osharko:


obj.go
```diff
@@ -217,6 +217,34 @@ func (O *Object) AsMap(recursive bool) (map[string]interface{}, error) {
...
+				return m, fmt.Errorf("%q: %w", a, err)
+			} else {
+				m[a] = r
+			}
+		}
+	}
+	return m, nil
+}
+
+func (o *Object) GetFromMap(m map[string]interface{}) {
```
tgulacsi, line 242: This horribly inefficient convenience function (just a for loop) should not be in the library.

obj.go
```diff
@@ -420,6 +448,33 @@ func (O ObjectCollection) Trim(n int) error {
 	return O.drv.checkExec(func() C.int { return C.dpiObject_trim(O.dpiObject, C.uint32_t(n)) })
 }
 
+func (O ObjectCollection) CollAsArrayMap() ([]map[string]interface{}, error) {
```
tgulacsi, line 451: What if the collection's elements are not objects?

obj.go
```diff
@@ -420,6 +448,33 @@ func (O ObjectCollection) Trim(n int) error {
...
+			continue
+		} else {
+			r, _ := value.(*Object).ObjAsMap()
+			array = append(array, r)
+		}
+	}
+	return array, nil
+}
+
+func (coll ObjectCollection) GetFromArrayMap(maps []map[string]interface{}) error {
```
tgulacsi, line 466: Why would sy create such a horrible data structure ([]map[string]interface{}) ?
Don't you know your object's structure beforehand?

tgulacsi (changes requested):
I've included (AsMap earlier) most of this in 57eabdd , PTAL.

obj.go
osharko: The godror.ObjectCollection could be called even with native types?
I create the collection starting from type, so i suppose it's always a custom type.

obj.go
osharko: In my case, often the BE just take the value from FE ad send it to DL.
So just converting json to map and viceversa, could be a best case scenario so the business logic will not be hold by the BE.

Btw even in other situation, i think it could be fine to send data from the json, and so work with the map/map array.

obj.go
osharko: it goes with GetFromArrayMap, to give the same option for the obj and for the collection.
I think it can be fine for converting the value then to a json (and i need it too, so taking into the library is good. Even i'll write it into my code).

Another point will be to convert to a scan, but i haven't figured it out yet.

osharko:
> I've included (AsMap earlier) most of this in [57eabdd](https://github.com/godror/godror/commit/57eabdde019421741b8008db36e18938feadf08e) , PTAL.

Seen, and i changed the refs of my code to that.
It's not working. I've a nested array of object, and the switch case identificate it's type as an Object

tgulacsi:
Why not store as json (needs Oracle 12, or even better 21)
________________________________

osharko:
It's a old legacy project on wich we're deploying nowdays.
It's a miracle that we can create a new backend for the new sections.
I'm not even on backend but on FE for this project, but i'm the tutor of most of these guys, so i'm working on be too for making the process quicker (write these call in java, with the 10+ layers that the app has, require also 4h for just a call, in go we need 15min using maps, and even lower time)

tgulacsi:
> > I've included (AsMap earlier) most of this in [57eabdd](https://github.com/godror/godror/commit/57eabdde019421741b8008db36e18938feadf08e) , PTAL.
> 
> Seen, and i changed the refs of my code to that. It's not working. I've a nested array of object, and the switch case identificate it's type as an Object
> 
> P.S. Just noticed you're working with data and not with godror.Object anymore. godror.Object take metadata of the object, I don't know if you could just modify the switch in data.go for a better working with godror.objectCollection, or you must keep working with godror.object.

A test case would help a lot.

osharko:
> A test case would help a lot.

This is from the discussion 204, is just above the other object declaration:

    TYPE DETAIL_REC IS OBJECT
    (
      resource_last_name      varchar2(100),
      resource_first_name     varchar2(100),
      resource_email          varchar2(100),
      owner_last_name      varchar2(100),
      owner_first_name     varchar2(100),
      owner_email          varchar2(100), 
      cardstatus varchar2(100),
      notes varchar2(2000),
      internal_level varchar2(100),
      card_history CARD_HISTORY_LIST
    );
    
    TYPE "CARD_HISTORY_LIST" IS table OF CARD_HISTORY_REC;
    
    TYPE "CARD_HISTORY_REC" IS OBJECT
    (
      card_status varchar2(200),
      notes varchar2(2000),
      event_user     varchar2(200),
      event_date    varchar2(30)
    )
    
    CREATE OR REPLACE procedure XXX_TEST.get_card_detail(p_userid           in varchar2 default null,
                                                p_address        in varchar2 default null,
                                                p_id_card          in number default null,
                                                p_language         in integer default 1,
                                                p_message          out varchar2,
                                                p_detail           out ETAIL_REC,
                                                p_code_application in varchar2 default 'BEC')

tgulacsi:
PTAL d89b41b

Wouldn't it be easier to convert to/from JSON directly, without the `map[string]interface{}` step?

osharko:
> PTAL [d89b41b](https://github.com/godror/godror/commit/d89b41b0a9924dd1e33abaa8df3b39973b006f20)
> 
> Wouldn't it be easier to convert to/from JSON directly, without the `map[string]interface{}` step?

I could say no for two main reason:
1) in some case the backend must be an orchestrator and made some other calls according to plsql response, so the map coul be easy to use, the JSON not;
2) json is about presentation layer, and the call to procedure is about business logic. So a controller will convert map to JSON and the Bl logic will handle the map.

The code seems to give the same response as mine.
Thank you.

For the other 2 function that I push, so that will not be merged into the plugin (the 2 func that convert from map to godror obj or Coll), right?

obj.go
tgulacsi: Ok, please rebase your code on top of the current master.
That invokes that CollAsArrayMap and ObjAsMap are not needed (we have AsMapSlice and AsMap).
For the remaining  GetFromArrayMap and GetFromMap:
1. rename them to FromMapSlice and FromMap
2. Put nice comment about what they do: Populate the Object/Collection from the map/[]map, and they're short, it's there mostly as an example.

tgulacsi:
Ok, please rebase your code on top of the current master.
That invokes that CollAsArrayMap and ObjAsMap are not needed (we have AsMapSlice and AsMap).
For the remaining GetFromArrayMap and GetFromMap:

rename them to FromMapSlice and FromMap
Put nice comment about what they do: Populate the Object/Collection from the map/[]map, and they're short, it's there mostly as an example.

osharko:
> FromMapSlice

I think I have quite finished rewriting it, because I noticed that I havent handled all the cases, but only the one of the object with flat type but not inner object/collection.

The error that now i'm facing is that the function Set, line 105 of obj.go, say me "godror.ObjectCollection: not supported", so i'm not able to add the  ObjectCollection to the object.

I pushed the new code, but it's not complete for now.

In the end, what would you suggest to solve the problem quoted as "This is horrendously inefficient" in the comment?

obj.go
```diff
@@ -345,6 +345,80 @@ func (O ObjectCollection) AsMapSlice(recursive bool) ([]map[string]interface{},
 	return m, nil
 }
 
+//Populate the Object starting from a map, according to the Object's Attributes.
```
tgulacsi, line 348: // FromMap populates ...

That's how godoc works.

obj.go
```diff
@@ -345,6 +345,80 @@ func (O ObjectCollection) AsMapSlice(recursive bool) ([]map[string]interface{},
...
+				O.Set(a, newO)
+			}
+		} else { //Plain type case
+			O.Set(a, v)
+		}
+	}
+	return nil
+}
+
+//Populate the ObjectCollection starting from a slice of map, according to the Collections's Attributes.
```
tgulacsi, line 399: // FromMapSlice populates ...

tgulacsi (changes requested):
Please add tests, too!
You can use GetObjectInObject or ObjectGetList, populate an Object, then check it's ToJSON produces the same json as should.

osharko:
> Please add tests, too! You can use GetObjectInObject or ObjectGetList, populate an Object, then check it's ToJSON produces the same json as should.

Added a test too.
The test will create a map, create a new object type and the object, and try to put the value from the map to the object.
Then the object will be reconverted to map and the two map will be evaluated.

This is the best test we can do, because only if the translation/detranslation really works fine, the test will be ok.

osharko:
> You can use GetObjectInObject or ObjectGetList, populate an Object, then check it's ToJSON produces the same json as should.

I haven't understood this line. Is it related to the test or related to this:
In the end, what would you suggest to solve the problem quoted as "This is horrendously inefficient" in the comment?

tgulacsi:
> > You can use GetObjectInObject or ObjectGetList, populate an Object, then check it's ToJSON produces the same json as should.
> 
> I haven't understood this line. Is it related to the test or related to this: In the end, what would you suggest to solve the problem quoted as "This is horrendously inefficient" in the comment?

It's related to the test. It may be easier (have less boilerplate) to just add some code to an existing test case, than write a new one, which has to create the DB types, too.

tgulacsi (approved):

osharko:
> It's related to the test. It may be easier (have less boilerplate) to just add some code to an existing test case, than write a new one, which has to create the DB types, too.

Ok, I see you merged but as I said previously, when I put the collection into the object with O.Set(attributeName, collection), the assignment fail. I don't know what is wrong, and I don't you how did you made something like this when you recreate the object given from oracle to the out object. But maybe I must change something and make it like you did.

tgulacsi:
> > It's related to the test. It may be easier (have less boilerplate) to just add some code to an existing test case, than write a new one, which has to create the DB types, too.
> 
> Ok, I see you merged but as I said previously, when I put the collection into the object with O.Set(attributeName, collection), the assignment fail. I don't know what is wrong, and I don't you how did you made something like this when you recreate the object given from oracle to the out object. But maybe I must change something and make it like you did.

`O.Set(a, ObjectCollection{...})` is allowed now on the current main (873800eab96b2cc53024c399c748c9a1eb15bc2c).

PR #208, opened 2022-01-06
Client character set override via a connection parameter

robstradling:
drv.go currently hardcodes UTF-8 (aka AL32UTF8) as the client character set.  I realize UTF-8 should be good enough for anyone these days, but unfortunately I'm developing a godror-based application that needs to use US7ASCII as the client character set instead of AL32UTF8.

My use case is that I have a >20yr old Oracle database that was naively created as US7ASCII but that holds 8-bit data that is predominantly UTF-8.  (This is possible because Oracle does not perform character set conversion when the client and server both claim to be using the same character set).  I realize that this set-up is most definitely "doing it wrong", but unfortunately I'm stuck with it for at least the foreseeable future.

If you'd prefer to define a custom environment variable for this instead of using NLS_LANG, I'd be fine with that.

### DIFF
--- a/drv.go
+++ b/drv.go
@@ -78,7 +78,6 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"os"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


tgulacsi:
The client encoding is fixed on UTF8 (AL32UTF8) because that's what Go's strings need.

Do I understand correctly, that you want to use US7ASCII, because that's the server's encoding, and you must avoid the charset conversion, because your DB holds invalid (not US7ASCII) data?

I don't want to obey NLS_LANG, because in 99% of the cases you want the charset conversion.
But an extra connection parameter could be used to lie to the server.

robstradling:
Yes, you've understood my need correctly.  I suspected you might not want to obey NLS_LANG.
An extra connection parameter to lie to the server SGTM.

robstradling:
Abandoned in favour of #209.

PR #213, opened 2022-01-25
Array of native values must be converted with AsSlice and not AsMapSlice

osharko:
Hi,
I found that the array of number or array of string was not converted.
The behavior was that array was initialized with the right length, but no value was inserted into it due the fact that the AsMapSlice was looking for an Object.
The correct way is to call  AsSlice instead of AsMapSlice.

### DIFF
--- a/obj.go
+++ b/obj.go
@@ -241,7 +241,11 @@ func (O *Object) AsMap(recursive bool) (map[string]interface{}, error) {
 				}
 				continue
 			}
-			if m[a], err = sub.Collection().AsMapSlice(recursive); err != nil {
+			if sub.ObjectType.CollectionOf.Attributes == nil {
+				if m[a], err = sub.Collection().AsSlice(nil); err != nil {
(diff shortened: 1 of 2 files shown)


tgulacsi:
Can you add some test?

osharko:
> 
As the previous times I have no way to test, but i created the test, it should work.
The concept is simple: A procedure return an object who hold an array of varchar, and then just make an AsMap.

osharko:
Will this commit be released asap?

tgulacsi:
Test/use it for a while, maybe you find more edge cases...

BTW I've fixed the test and moved it to `z_plsql_types_test.go`.

osharko:
> Test/use it for a while, maybe you find more edge cases...
> 
> BTW I've fixed the test and moved it to `z_plsql_types_test.go`.

Ok, how do i add it to the go mod?

osharko:
> Will this commit be released asap?
Oh right, ty
go get github.com/godror/godror@main

PR #264, opened 2022-11-21
fix dsn password exists special characters

xjustloveux:
If dsn password exists special characters, then password need in double quotes.

ex:
password = pa@ss
dsn = user/"pa@ss"@sid

### DIFF
--- a/dsn/dsn.go
+++ b/dsn/dsn.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


dsn/dsn.go
```diff
@@ -776,6 +776,14 @@ func parseUserPassw(dataSourceName string) (user, passw, connectString string) {
 		return "", "", dataSourceName + unquote(extra)
 	}
 
+	// username/"password"
+	if strings.Index(dataSourceName, "/\"") > 0 {
+		if si := strings.Index(dataSourceName, "\""); si > 0 {
+			li := strings.Index(dataSourceName[si+1:], "\"") + si
```
tgulacsi, line 782: What if no other " is there? 
Please handle all kind of errors!

tgulacsi (changes requested):

dsn/dsn.go
xjustloveux: I fix it better and add test, but it only check user/"pass" and user/"pass"@sid.
Other dsn don't need this check.

xjustloveux: I don't think I need to confirm the case where there is only one double quote input.
Because the dsn was originally wrong, it will return a connection failed error in the subsequent code.
Oracle password doesn't allow double quotes.
[https://docs.oracle.com/cd/E11223_01/doc.910/e11197/app_special_char.htm#MCMAD416](url)
My code is only used to confirm that when there are special characters in the password, the password in double quotes can be used to connect success.

xjustloveux:
Thanks for your reply.

Issue #219, opened 2022-02-19
"ConnectionParams.OnInit" sometimes does not take effect under concurrent conditions


cartersz:
**Describe the bug**
The code that is "ALTER SESSION SET NLS_TIMESTAMP_FORMAT YYYY-MM-DD HH24.MI.SS.FF" used to change default session parameter "NLS_TIMESTAMP_FORMAT" with "ConnectionParams.OnInit" of configuration item, but sometimes it does not work, some data can be inserted with correct session timestamp formate, but some data cannot be inserted with fault session timestamp formate. And, it is the same problem using "ConnectionParams.SetSessionParamOnInit" and "ConnectionParams.OnInitStmts".

**To Reproduce**
The following code which can reproduce the problem.
```go
Table T1:
create table T1
(
    C1 NUMBER not null,
    C2 TIMESTAMP(6),
    C3 DATE,
    C4 TIMESTAMP(0),
    C5 DATE
)

create unique index T1_C1_UINDEX
    on T1 (C1)

import (
	"context"
	"database/sql"
	"database/sql/driver"
	"fmt"
	"github.com/godror/godror"
	"github.com/pingcap/errors"
	"time"

	_ "github.com/go-sql-driver/mysql"
)

   func main() {
	insertTime()
   }

   func insertTime() {
	dbConnectionPool := buildNewOracleConnectionTimeInsert()
	_, err := dbConnectionPool.Exec("truncate table USER.T1")
	if err != nil {
		fmt.Println("truncate table error", err)
	}
	for i := 0; i < 100*100; i++ {
		go doExec(i, dbConnectionPool)
	}
	time.Sleep(60 * time.Second)
   }

   func doExec(i int, db *sql.DB){
	tx,_ := db.Begin()
	sql := "INSERT INTO USER.T1 (C1, C2, C3, C4, C5) VALUES (:1, :2, :3, :4, :5)"
	args := make([]interface{}, 5)
	args[0] = 11 + i
	args[1] = "2022-01-18 11:04:49.123456"
	args[2] = "2022-01-18"
	args[3] = "2022-01-18 11:04:49"
	args[4] = "2022-01-18 11:04:49"
	_, err := tx.Exec(sql, args...)
	if err != nil {
		fmt.Println(fmt.Sprintf("#########insert exec error: i = %d, %s" , i, showSessionParam(tx)), err)
	}else {
[...]

Issue #218, opened 2022-02-18
AlterSession 'Current_Schema' is not honoured past first session


travellingkiwi:
When creating a connection to an Oracle database (12c in case it matters) using godror (golang 1.17) the first query works fine. However subsequent queries result in 'ORA-00942 table or view does not exist' for a simple SQL query when the schema is not specified in the query

e.g.
select * from my_table;

works once and then fails. While
select * from my_schema.my_table;

continues to work

I have tried setting the current_schema via directly executing 'alter session set current_schema=my_schema' and also specifying the current_schema in the connection parameter AlterSession array. Both methods result in a single successful run of the query followed by continual failures with ORA-00942

sudarshan12s:
> 

I don't see any such issues in my 12.1 setup. Can you share more details on test case with table , column types ,...  , preferably a minimal program, if possible.

travellingkiwi:
I've narrowed it down slightly. It's actually the first query that only succeeds. It's that only statements prepared on the first connection succeed. Because only the first session in the pool gets the current_schema set to a non-default value.

When I connect I create the connection via

`
var P godror.ConnectionParams
P.username, P.Password="myuser", godror.NewPassword("mypass")
P.ConnectString="mydb"
P.AlterSession=[][2]string{{"CURRENT_schema", "my_schema_owner"}}

self.pool=sql.OpenDB(godror.NewConnector(P))
`

And then calling

`
self.pool.Prepare("some_sql")
`

works fine... For the first prepare... But subsequent prepares fail with ORA-00942 (The Sql doesn't refer to the table by schema.table_name, just by table_name and relying on the current_schema)

When I look at the v$session for the database after the failures start, I can see that I have 3-4 sessions. The first session lists the schemaname correctly as "my_schema_owner". The subsequent sessions for the pool (identified as being subsequent by them having a higher client port) have the connection user as the schemaname...
So it looks like only the first session of the pool is getting the AlterSession run...

Is there something special I should be doing to have all sessions in the pool execute the AlterSchema parameters?

(I've also tried leaving AlterSession empty. And just running 

`
self.pool.Exec("alter session set current_schema='my_schema_owner'" 
`

But it has exactly the same issue.

travellingkiwi:
Edited the title as it appears the root cause is that only the first session of the pool has AlterSession run for it.

tgulacsi:
PTAL e575822 with setting logger with `godror.SetLog` to see what and when is executed.
[...]

PR #266, opened 2022-12-01
Provide Info about new connection to OnInit method

stanowawi:
Intention for new Feature:
The ODPI Driver provides information if the connection has been newly created, or it is taken from the connection pool ( see doc for dpiConnCreateParams.outNewSession) 
It is really useful to have this information by the OnInit method, to be able to differentiate between  connection from the pool or new connection.
Change:
- get and forward the "outNewSession" info as a return value by createConn & acquireConn  methods (newConnectionCreated true = new connection / false = pooled connection)
- Forward this info to the OnInit Method as a  new context tag. This solution with the context tag doesn't change OnInit signature = is  backwards compatible 
- Create new dsn.OnInitNewConnectionType / OnInitNewConnectionKey to provide unique context key for all users. 

 The whole change is backwards compatible.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1107,13 +1107,13 @@ func (c *conn) ResetSession(ctx context.Context) error {
 		// Just release
 		_ = c.closeNotLocking()
 	}
-	dpiConn, newConnectionCreated, err := c.drv.acquireConn(pool, P)
+	dpiConn, isNew, err := c.drv.acquireConn(pool, P)
 	c.mu.Unlock()
 	if err != nil {
 		return fmt.Errorf("%v: %w", err, driver.ErrBadConn)
 	}
 	c.dpiConn = dpiConn
-	ctx = context.WithValue(ctx, dsn.OnInitNewConnectionKey, newConnectionCreated)
+	ctx = context.WithValue(ctx, dsn.OnInitNewConnectionKey, isNew)
 	return c.init(ctx, P.OnInit)
 }
 
--- a/conn.go
+++ b/conn.go
@@ -501,7 +501,7 @@ func (c *conn) init(ctx context.Context, isNew bool, onInit func(ctx context.Con
 		logger.Log("msg", "connection initialized", "conn", c, "haveOnInit", onInit != nil)
 	}
 
-	if c.params.CommonParams.OnInitNewCon && !isNew {
+	if c.params.CommonParams.InitOnNewConn && !isNew {
 		return nil
 	}
 
--- a/conn.go
+++ b/conn.go
@@ -1113,7 +1113,7 @@ func (c *conn) ResetSession(ctx context.Context) error {
 		return fmt.Errorf("%v: %w", err, driver.ErrBadConn)
 	}
 	c.dpiConn = dpiConn
-	ctx = context.WithValue(ctx, dsn.OnInitNewConnectionKey, isNew)
+	ctx = context.WithValue(ctx, dsn.isNewCtxKey, isNew)
 	return c.init(ctx, P.OnInit)
 }
 
--- a/conn.go
+++ b/conn.go
@@ -1107,13 +1107,13 @@ func (c *conn) ResetSession(ctx context.Context) error {
 		// Just release
 		_ = c.closeNotLocking()
 	}
-	dpiConn, err := c.drv.acquireConn(pool, P)
+	dpiConn, newConnectionCreated, err := c.drv.acquireConn(pool, P)
 	c.mu.Unlock()
 	if err != nil {
 		return fmt.Errorf("%v: %w", err, driver.ErrBadConn)
 	}
 	c.dpiConn = dpiConn
-
+	ctx = context.WithValue(ctx, dsn.OnInitNewConnectionKey, newConnectionCreated)
 	return c.init(ctx, P.OnInit)
 }
 
--- a/conn.go
+++ b/conn.go
@@ -487,7 +487,7 @@ func (c *conn) ServerVersion() (VersionInfo, error) {
 	return c.Server, nil
 }
 
-func (c *conn) init(ctx context.Context, onInit func(ctx context.Context, conn driver.ConnPrepareContext) error) error {
+func (c *conn) init(ctx context.Context, isNew bool, onInit func(ctx context.Context, conn driver.ConnPrepareContext) error) error {
 	c.released = false
 	logger := ctxGetLog(ctx)
 	if logger != nil {
@@ -500,6 +500,11 @@ func (c *conn) init(ctx context.Context, onInit func(ctx context.Context, conn d
 	if logger != nil {
 		logger.Log("msg", "connection initialized", "conn", c, "haveOnInit", onInit != nil)
 	}
+
+	if c.params.CommonParams.OnInitNewCon && !isNew {
+		return nil
+	}
+
 	return onInit(ctx, c)
 }
 
@@ -1113,8 +1118,8 @@ func (c *conn) ResetSession(ctx context.Context) error {
 		return fmt.Errorf("%v: %w", err, driver.ErrBadConn)
 	}
 	c.dpiConn = dpiConn
-	ctx = context.WithValue(ctx, dsn.isNewCtxKey, isNew)
-	return c.init(ctx, P.OnInit)
+
+	return c.init(ctx, isNew, P.OnInit)
 }
 
 // Validator may be implemented by Conn to allow drivers to
--- a/drv.go
+++ b/drv.go
@@ -613,11 +613,9 @@ func (d *drv) createConnFromParams(ctx context.Context, P dsn.ConnectionParams) 
 	if onInit == nil {
 		return conn, err
 	}
-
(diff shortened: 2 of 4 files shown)


tgulacsi:
See defb84496261a07de2829aa84e7ec66264e24065 and #218, #219 - that outNewSession lied reproducibly.

stanowawi:
I'm trying to reproduce the issue with the test out of #219 and my modification like this:
```
oraDSN.OnInit = func(ctx context.Context, conn driver.ConnPrepareContext) error {

newConFromOCI := ctx.Value(dsn.OnInitNewConnectionKey).(bool)
if newConFromOCI {
	qry := "ALTER SESSION SET NLS_DATE_FORMAT='YYYY-MM-DD HH24.MI.SS' NLS_TIMESTAMP_FORMAT='YYYY-MM-DD HH24.MI.SS.FF' NLS_TIMESTAMP_TZ_FORMAT='YYYY-MM-DD HH24.MI.SS.FF TZR' NLS_TIME_FORMAT='HH24.MI.SS.FF' NLS_TIME_TZ_FORMAT='HH24.MI.SS.FF TZR'"
	st, err := conn.PrepareContext(ctx, qry)
	if err == nil {
		_, err = st.(driver.StmtExecContext).ExecContext(ctx, nil)
		st.Close()
	}
....
```
I run it lot of times with the instantclient_19_16, instantclient_12_2 and instantclient_21_8, but. I can see the logic works, but I'm getting no issue :(
Really strange, do you have some recommendation?

conn.go
```diff
@@ -1107,13 +1107,13 @@ func (c *conn) ResetSession(ctx context.Context) error {
 		// Just release
 		_ = c.closeNotLocking()
 	}
-	dpiConn, err := c.drv.acquireConn(pool, P)
+	dpiConn, newConnectionCreated, err := c.drv.acquireConn(pool, P)
```
tgulacsi, line 1110: User shorter names, like "isNew"

conn.go
```diff
@@ -1107,13 +1107,13 @@ func (c *conn) ResetSession(ctx context.Context) error {
...
 	c.mu.Unlock()
 	if err != nil {
 		return fmt.Errorf("%v: %w", err, driver.ErrBadConn)
 	}
 	c.dpiConn = dpiConn
-
+	ctx = context.WithValue(ctx, dsn.OnInitNewConnectionKey, newConnectionCreated)
```
tgulacsi, line 1116: Why do we put this information into the context?

drv.go
```diff
@@ -369,15 +369,16 @@ func (d *drv) initCommonCreateParams(P *C.dpiCommonCreateParams, enableEvents bo
...
+// second return value: true = connection is new / false = connection is from pool
+func (d *drv) createConn(pool *connPool, P commonAndConnParams) (*conn, bool, error) {
 	// initialize driver, if necessary
 	if err := d.init(P.ConfigDir, P.LibDir); err != nil {
-		return nil, err
+		return nil, false, err
 	}
 
-	dc, err := d.acquireConn(pool, P)
+	dc, newConnectionCreated, err := d.acquireConn(pool, P)
```
tgulacsi, line 379: isNew

drv.go
```diff
@@ -396,10 +397,10 @@ func (d *drv) createConn(pool *connPool, P commonAndConnParams) (*conn, error) {
 			c.params.Username = pool.params.Username
 		}
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), nvlD(c.params.WaitTimeout, time.Minute))
+	ctx, cancel := context.WithTimeout(context.WithValue(context.Background(), dsn.OnInitNewConnectionKey, newConnectionCreated), nvlD(c.params.WaitTimeout, time.Minute))
```
tgulacsi, line 400: why?

drv.go
```diff
@@ -571,13 +572,15 @@ func (d *drv) acquireConn(pool *connPool, P commonAndConnParams) (*C.dpiConn, er
...
+			return nil, false, fmt.Errorf("pool=%p stats=%s params=%+v: %w",
 				pool.dpiPool, stats, connCreateParams, err)
 		}
-		return nil, fmt.Errorf("user=%q standalone params=%+v: %w",
+		return nil, false, fmt.Errorf("user=%q standalone params=%+v: %w",
 			username, connCreateParams, err)
 	}
-	return dc, nil
+	//use the information from ODPI driver if new connection has been created or it is only pooled
+	newConnctionCreated := connCreateParams.outNewSession == 1
```
tgulacsi, line 582: isNew

drv.go
```diff
@@ -597,15 +600,15 @@ func (d *drv) createConnFromParams(ctx context.Context, P dsn.ConnectionParams)
 			return nil, err
 		}
 	}
-	conn, err := d.createConn(pool, commonAndConnParams{CommonParams: P.CommonParams, ConnParams: P.ConnParams})
+	conn, newConnectionCreated, err := d.createConn(pool, commonAndConnParams{CommonParams: P.CommonParams, ConnParams: P.ConnParams})
```
tgulacsi, line 603: isNew

drv.go
```diff
@@ -597,15 +600,15 @@ func (d *drv) createConnFromParams(ctx context.Context, P dsn.ConnectionParams)
...
 	if err != nil {
 		return conn, err
 	}
 	onInit := getOnInit(&conn.params.CommonParams)
 	if onInit == nil {
 		return conn, err
 	}
-	ctx, cancel := context.WithTimeout(ctx, nvlD(conn.params.WaitTimeout, time.Minute))
+	ctx, cancel := context.WithTimeout(context.WithValue(ctx, dsn.OnInitNewConnectionKey, newConnectionCreated), nvlD(conn.params.WaitTimeout, time.Minute))
```
tgulacsi, line 611: ?

dsn/dsn.go
```diff
@@ -45,6 +45,13 @@ const (
 	DefaultStandaloneConnection = false
 )
 
+type OnInitNewConnectionType string
```
tgulacsi, line 48: Why do we need this at all?
If we do, use 
type isNewCtxKey struct{}

tgulacsi (changes requested):
Ok, let's try it, but make it OPT-IN with an Option!

stanowawi:
Hello Tamás.
Do you mean to change OnInit method and add optional parameter? This is I wanted to avoid, because all projects, which use it, will brake like this:  https://go.dev/play/p/j62gJWH1Xl5
I can see following options:
1. change OnInit signature (add new optional param) = not compatible
2. provide the "isNew" information over  Context, as I did = not nice but compatibility is not broken
3. introduce new OnInitNew with isNew Information method and make a OnInit deprecated

Do you want to take option 1. ?

conn.go
stanowawi: See comment bellow. I don't want to change OnInit method and cause compile errors in other projects, the I use Context to provide info.

drv.go
stanowawi: yes, you are right I'll change it

conn.go
tgulacsi: Why would OnInit need this info?
It should be called iff it is a new connection.

dsn/dsn.go
tgulacsi: Yup, 
```type isNewCtxKey struct{}```
is a new dedicated type for it. Do NOT use a string !

File: conn.go (hunk 2 above)
Comment (stanowawi) [reply conn.go]: I reworked according your requirements. Parameter = OnInitNewCon if true , then OnInit is called only by the new connection.
Please check.

File: dsn/dsn.go [hunk 9]
```diff
@@ -56,6 +56,8 @@ type CommonParams struct {
 	OnInit func(context.Context, driver.ConnPrepareContext) error
 	// OnInitStmts are executed on session init, iff OnInit is nil.
 	OnInitStmts []string
+	// true: OnInit will be called only by the new session / false: OnInit will called by new or pooled connection
+	OnInitNewCon bool
```
tgulacsi, line 60: Where does this get a value?

drv.go
```diff
@@ -597,20 +600,22 @@ func (d *drv) createConnFromParams(ctx context.Context, P dsn.ConnectionParams)
...
 	if onInit == nil {
 		return conn, err
 	}
-	ctx, cancel := context.WithTimeout(ctx, nvlD(conn.params.WaitTimeout, time.Minute))
-	err = onInit(ctx, conn)
-	cancel()
-	if err != nil {
-		conn.Close()
-		return nil, err
+	if !P.CommonParams.OnInitNewCon || isNew {
```
tgulacsi, line 611: Turn this around:
```
if P.CommonParams.OnInitNewCon && !isNew {
    return conn, nil
}
```

tgulacsi (changes requested):

tgulacsi (changes requested):
Sorry, one last stylistic change, please:
CommonParams.InitOnNewConn
and call the new param
"initOnNewConnection"
Thanks!

stanowawi:
no problem 
the rename is done, please check. Thank you too.

Issue #295, opened 2023-06-12
Serializable isolation level not honored?


minollo:
I have a simple scenario where I do:

- txn1 <- db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelReadCommitted, ReadOnly: false}
- txn2 <- db.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable, ReadOnly: true}
- readSomeData(txn2): I get no results
- writeSomeData(txn1); I don't "commit" yet
- readSomeData(txn2): I get no results - good
- txn1.Commit()
- readSomeData(txn2): I get data written in txn1

I wouldn't expect that to happen; "serializable" should make sure txn2 won't see any external commits which happen after the transaction has started...
What am I missing?

minollo:
It works as expected if in conn.BeginTx() I also execute...
  ALTER SESSION SET ISOLATION_LEVEL=SERIALIZABLE
...when the serializable level is requested; not sure why I would to do that - and whether that has some drawbacks.

tgulacsi:
PTAL TestLevelSerializable in z_test.go at [7d9f260](https://github.com/godror/godror/commit/7d9f2609636f380bbbe3f36362a818f7bca95640) - it works as expected.

You can check what happens by setting DPI_DEBUG_LEVEL=92 and search for "SQL SET TRANSACTION ISOLATION LEVEL" in the output.

tgulacsi:
READ ONLY or SERIALIZABLE ?
For an application standpoint, READ ONLY is more strict (does not allow modifications),
for DB level, SERIALIZABLE is the harder.

So the driver sets READ ONLY when you specify it, and SERIALIZABLE when you set {sql.LevelSerializable, ReadOnly: *false*}.

TL;DR; do not set ReadOnly when you don't mean it!

minollo:
How does that test verifies it? You should have a tx2.commit() after the tx1.Exec() operation; but you don't - you eventually just rollback tx1 and tx2.
If you add tx1.Commit() after the "insert into" execution, I expect you will see the same issue I'm seeing.

minollo:
In fact, if you add tx1.Commit() (ignore the lack of error checking):
...
	if _, err := tx1.ExecContext(ctx, "INSERT INTO "+tbl+" (id) VALUES (:1)", 3); err != nil {
		t.Fatal("insert1:", err)
	}
	if got := C(tx1); got != 1 {
[...]

PR #296, opened 2023-06-13
Related to issue https://github.com/godror/godror/issues/295

minollo:
- Fixed TestLevelSerializable() to perform Commit on tx1 before checking that tx2 is not affected by it.
- Switched to SESSION level isolation level when starting a transaction, as transaction level isolation level doesn't seem to be honored.


tgulacsi (changes requested):
I think the current main contains this logic.

PR #351, opened 2024-11-19
Nil pointer bugfix and 2 options

bauchpinsel:
Hi @tgulacsi,

I want so submit two minor changes that I need for my current project:

WarningAsError option mimics the behaviour of SQL*PLUS which gives feedback when a PL/SQL object was created successfully, but had compilation errors.

NoRetry option to prevent re-execution on certain errors. Although I think the re-try is generally a good call, re-execution might cause side-effects. For example in the following statement procedure1 and procedure2 will be re-executed which in my case must not happen.
```
BEGIN
  procedure1;
  procedure2;
  procedure3; -- ORA-04068 happens here
END;
```
Please let me know what you think and whether I should change anything.

### DIFF
--- a/drv.go
+++ b/drv.go
@@ -279,6 +279,14 @@ func (d *drv) checkExecNoLOT(f func() C.int) error {
 	return d.getError()
 }
 
+func (d *drv) checkExecWithWarning(f func() C.int) error {
+	runtime.LockOSThread()
+	_ = f()
+	err := d.getError()
+	runtime.UnlockOSThread()
+	return err
+}
+
 func (d *drv) init(configDir, libDir string) error {
 	d.mu.RLock()
 	ok := d.pools != nil && d.timezones != nil && d.dpiContext != nil
--- a/z_test.go
+++ b/z_test.go
@@ -5222,3 +5222,50 @@ func TestStandaloneVsPool(t *testing.T) {
 		db.Close()
 	}
 }
+
+func TestWarningAsError(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	const query = `
+		DECLARE
+			ora_24344 EXCEPTION;
+			PRAGMA EXCEPTION_INIT(ora_24344, -24344);
+		BEGIN
+			RAISE ora_24344;
+		END;`
+	args := []any{godror.WarningAsError()}
+	for _, tC := range []struct {
+		Qry     string
+		Code    int
+		Warning bool
+	}{
+		// Some syntax error
+		{Qry: "SELECT 1 FROM not-exist", Code: 903},
+		// No error - warning should not return the previous error
+		{Qry: "SELECT 1 FROM DUAL", Warning: true},
+		// Not a syntax error, but a compile error - just a warning
+		{Qry: query},
+		// Proper error
+		{Qry: `BEGIN raise_application_error(-20001, 'test'); END;`, Code: 20001},
+		// Warning as error
+		{Qry: query, Code: 24344, Warning: true},
+	} {
+		args = args[:0]
+		if tC.Warning {
+			args = args[:1]
+		}
+		var ec interface{ Code() int }
+		_, err := testDb.ExecContext(ctx, tC.Qry, args...)
+		if tC.Code == 0 && err != nil {
+			t.Errorf("got %+v, wanted no err", err)
+		} else if tC.Code != 0 {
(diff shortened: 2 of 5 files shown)


tgulacsi (commented):
First of all, thanks for this nice PR'

NoRetry: I'd love to have no such option, and your procedures should not have side effect (till COMMIT), but I understand that life is not as nice for legacy maintainers as we'd like. 
So I accept the need.

WarningAsError: my first thought was to add ths "warningAsError bool" argument to checkExec - but as we use checkExec a lot, that'd induce a big code churn without much gain. So it is ok this way (separate checkExecWithWarning function).

logger!=nil: thanks!

drv.go
```diff
@@ -279,6 +279,14 @@ func (d *drv) checkExecNoLOT(f func() C.int) error {
 	return d.getError()
 }
 
+func (d *drv) checkExecWithWarning(f func() C.int) error {
+	runtime.LockOSThread()
+	_ = f()
+	err := d.getError()
```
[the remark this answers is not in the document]
bauchpinsel: I added a TestWarningAsError in z_test.go to verify the behaviour. I just added another error in the test case to make sure the exceptions do not mix.

bauchpinsel:
"TestWarningAsError". My first approach was to add a boolean value. But as you said, that was a big change, and I thought you might not like it because it interferes so much with the rest of your codebase. This was the least impactful change I could make without it being a global flag, which I didn't like. 

Regarding the NoRetry option. I fully agree with you, but I want to give you some context, so that you better understand where I am coming from. We have over 50 active developers working on our PL/SQL legacy codebase and we have thousands of packages and millions of lines of code. I am working on a new installation tool for our software, so this might not be the most common way for using Godror.
If you try to install a new version of this software with all of its DDL, DML and Package scripts (which make up a new version), where all the DDL, DML and Packages all are cross referencing each other, you'll get a lot of invalidation errors. That's expected. But the hidden control flow that code is re-executed is something that is hard to deal with in my case. It helps to call dbms_session.reset_package a lot, but it doesn't help if our developers are getting really clever ;)

z_test.go
```diff
@@ -5222,3 +5222,30 @@ func TestStandaloneVsPool(t *testing.T) {
 		db.Close()
 	}
 }
+
+func TestWarningAsError(t *testing.T) {
```
tgulacsi, line 5226: Please use a table test, like
```
func TestWarningAsError(t *testing.T) {
        ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
        defer cancel()
        const query = `
                DECLARE
                        ora_24344 EXCEPTION;
                        PRAGMA EXCEPTION_INIT(ora_24344, -24344);
                BEGIN
                        RAISE ora_24344;
                END;`
        args := []any{godror.WarningAsError()}
        for _, tC := range []struct {
                Qry     string
                Code    int
                Warning bool
        }{
                // Some syntax error
                {Qry: "SELECT 1 FROM not-exist", Code: 903},
                // No error - warning should not return the previous error
                {Qry: "SELECT 1 FROM DUAL", Warning: true},
                // Not a syntax error, but a compile error - just a warning
                {Qry: query},
                // Proper error
                {Qry: `BEGIN raise_application_error(-20001, 'test'); END;`, Code: 20001},
                // Warning as error
                {Qry: query, Code: 24344, Warning: true},
        } {
                args = args[:0]
                if tC.Warning {
                        args = args[:1]
                }
                var ec interface{ Code() int }
                _, err := testDb.ExecContext(ctx, tC.Qry, args...)
                if tC.Code == 0 && err != nil {
                        t.Errorf("got %+v, wanted no err", err)
                } else if tC.Code != 0 {
                        if err == nil {
                                t.Errorf("wanted error for %q", tC.Qry)
                        } else if !errors.As(err, &ec) {
                                t.Errorf("wanted OraErr, got %#v", err)
                        } else if got := ec.Code(); got != tC.Code {
                                t.Errorf("wanted ORA-%d, got %d", tC.Code, got)
                        }
                }
        }

}
```

tgulacsi (changes requested):

z_test.go
bauchpinsel: I rebased the commit with your recommendation.

bauchpinsel: Sorry in case I missed an obvious Github workflow. This is my first pull request. I just rebased your change in my local repo and force pushed. I hope that's ok.

tgulacsi (approved):

PR #364, opened 2025-03-17
- Adding support for Vector and Sparse Vector

sudarshan12s:
### Description

Added 23ai [vector](https://docs.oracle.com/en/database/oracle/oracle-database/23/vecse/create-tables-using-vector-data-type.html) support. 

Dense  vector (`VECTOR`) and Sparse vector (`VECTOR(dims, format, SPARSE)`) support has been added. 

A new type, Vector is added to encapsulate Vector and Sparse Vector data.

```
type Vector struct {
	Dimensions uint32      // Total dimensions of the vector.
	Indices    []uint32    // Indices of non-zero values (sparse format).
	Values     interface{} // Non-zero values (sparse format) or all values (dense format).
	IsSparse   bool        // Flag to detect if it's a sparse vector
}
```

New Column MetaData of Vector is added
- VectorFormat -> It represents one of the formats: float32, float64, int8, uint8 (Binary vector)
- VectorFlags -> Indicates if Vector is having flexible dimensions, Sparse
- VectorDimensions -> Number of Dimensions.
 
### Example Usage:

```
CREATE TABLE tbl (id NUMBER(6), embedding Vector)
stmt, err := conn.PrepareContext(ctx,"INSERT INTO tbl (id, embedding) VALUES (:1, :2)")
if _, err = stmt.ExecContext(ctx, 1, &godror.Vector{Values: []float32{1.1, 2.2, 3.3}}); err != nil {
    log.Fatalf("insert failed: %v", err)
}
```

**Bind examples:**

**- Dense Vector** 
godror.Vector{Values: []float32{1.1, 2.2, 3.3}} // float32 storage format
&godror.Vector{Values: []float64{1.1, 2.2, 3.3}} // float64 storage format , passing address of godror.Vector

**- Sparse Vector**
- {Values: []int8{-1, 4, -7}, Indices: []uint32{1, 2, 3}, Dimensions: 4}, // int8 storage format Sparse Vector
-  {Values: []float32{-1, 4, -7}, Indices: []uint32{1, 2, 3}, Dimensions: 4}, // float32 storage format Sparse Vector

On reading from DB, we get the godror.Vector type returned.

```
row := conn.QueryRowContext(ctx, fmt.Sprintf(`SELECT embedding from tbl`))
var outVal godror.Vector
err = row.Scan(&outVal)
```
	

**Note:** 
1. generics like below is explored but as its not available until go 1.18 , it is not considered.
```
 // Storage Format supported types
type Format interface {
	int8 | float32 | float64 | uint8
}

// Vector holds the embedding VECTOR column starting from 23ai.
type Vector[T Format] struct {
	Indices    []uint32 // Indices of non-zero values (sparse format)
	Dimensions uint32   // Total dimensions of the vector (can be set explicitly)
	Values     []T      // Non-zero values (sparse format) or all values (dense format)
	IsSparse   bool     // Flag to detect if it's a sparse vector
}

```
2. In SetVectorValue function, when I wanted to pass go slice to C pointers, It works fine for float32, float64 but did not work for uint32 slices. Hence  I am doing malloc and free for uint32_t, uint8 and int8 slices.

### DIFF
--- a/vector.go
+++ b/vector.go
@@ -37,6 +37,12 @@ type Vector struct {
 
 // SetVectorValue converts a Go `Vector` into a godror data type.
 func SetVectorValue(c *conn, v *Vector, data *C.dpiData) error {
+	if vi, err := c.ClientVersion(); err != nil {
+		return err
+	} else if vi.Version < 23 || vi.Version == 23 && vi.Release < 7 {
+		return fmt.Errorf("clientVersion is old (%s, sparse vector needs at least 23.7)", vi.String())
+	}
+
 	var vectorInfo C.dpiVectorInfo
 	var valuesPtr unsafe.Pointer
 	var format C.uint8_t
--- a/vector.go
+++ b/vector.go
@@ -57,24 +57,12 @@ func SetVectorValue(c *conn, v *Vector, data *C.dpiData) error {
 	case []int8:
 		numDims, format, dimensionSize = len(values), C.DPI_VECTOR_FORMAT_INT8, 1
 		if numDims > 0 {
-			ptr := (*C.int8_t)(C.malloc(C.size_t(numDims)))
-			defer C.free(unsafe.Pointer(ptr))
-			cArray := unsafe.Slice((*int8)(unsafe.Pointer(ptr)), numDims)
-			for i, v := range values {
-				cArray[i] = int8(v)
-			}
-			valuesPtr = unsafe.Pointer(ptr)
+			valuesPtr = unsafe.Pointer(&values[0])
 		}
 	case []uint8:
 		numDims, format, dimensionSize = len(values), C.DPI_VECTOR_FORMAT_BINARY, 1
 		if numDims > 0 {
-			ptr := (*C.uint8_t)(C.malloc(C.size_t(numDims)))
-			defer C.free(unsafe.Pointer(ptr))
-			cArray := unsafe.Slice((*uint8)(unsafe.Pointer(ptr)), numDims)
-			for i, v := range values {
-				cArray[i] = uint8(v)
-			}
-			valuesPtr = unsafe.Pointer(ptr)
+			valuesPtr = unsafe.Pointer(&values[0])
 		}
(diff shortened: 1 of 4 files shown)


vector.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	case []float64:
+		numDims, format, dimensionSize = len(values), C.DPI_VECTOR_FORMAT_FLOAT64, 8
+		if numDims > 0 {
+			valuesPtr = unsafe.Pointer(&values[0])
+		}
+	case []int8:
+		numDims, format, dimensionSize = len(values), C.DPI_VECTOR_FORMAT_INT8, 1
+		if numDims > 0 {
+			ptr := (*C.int8_t)(C.malloc(C.size_t(numDims)))
```
tgulacsi, line 60: What's the error here (why do we need malloc, why not just an unsafe.Pointer(&values[0])?

tgulacsi (commented):
I cannot test this (I'm stuck on 19.23 and lost my access to the Oracle Autonomous Database as the authentication methods constantly change), but seems nice, thanks!

anthony-tuininga:
> (ORA-00000: DPI-1049: symbol OCIVectorFromSparseArray not found in OCI library)

I have just pushed changes to improve that error message. The error message you will get now is:

```
DPI-1050: Oracle Client library is at version 23.4 but version 23.7 or higher is needed
```

tgulacsi:
Add it now, will remove it on next OFPI-C release.

sudarshan12s:
> Add it now, will remove it on next OFPI-C release.

made suggested changes. 
with IC 23.6, if user binds sparse vector, following error is thrown:
`clientVersion is old (23.6.0, sparse vector needs at least 23.7)
`
With IC <23.4 (here 21.3 is used) , following error is thrown
`ORA-00000: DPI-1050: Oracle Client library is at version 21.3 but version 23.4 or higher is needed
`

tgulacsi (approved):

PR #373, opened 2025-06-03
Batch enhancements & coverage improvements

lukius:
# Summary

This PR enhances the batch processing functionality by adding a new method to access SQL execution results and provides comprehensive test coverage for batch operations.

## Problem statement

The current batch implementation only provides basic error reporting through the `Flush()` method, which returns a simple error. Applications that need access to execution details (e.g. `RowsAffected()`) have no way to retrieve the `sql.Result` from batch operations, limiting their ability to validate execution success or gather execution metrics.

### New `FlushWithResult` method
- Added `FlushWithResult(ctx context.Context) (sql.Result, error)` method that executes the batch and returns the `sql.Result` for applications needing access to execution details
- Refactored `Flush()` to use `FlushWithResult()` internally, eliminating code duplication while maintaining backward compatibility
- Maintains the same execution logic but provides flexibility for different use cases

### Improved error handling
- Enhanced error handling in both `Add()` and `FlushWithResult()` methods with better error context
- Added proper error checking for `RowsAffected()` calls to catch potential driver-level issues

### Test coverage
Added 10 new test functions covering various scenarios:
- Basic batch operations and error handling
- Empty batch flush behavior
- Nil value handling in batches
- Auto-flush functionality
- Concurrent batch usage
- `FlushWithResult` method functionality
- `FlushWithResult` error handling
- `FlushWithResult` empty batch behavior

### Backward compatibility
- The `Flush()` method signature remains unchanged
- Enhanced functionality is opt-in through the new `FlushWithResult()` method

### New method
```go
// FlushWithResult executes the statement, clears the storage, and returns the result.
func (b *Batch) FlushWithResult(ctx context.Context) (sql.Result, error)
```

### Usage examples
```go
// For applications that need access to sql.Result
result, err := batch.FlushWithResult(ctx)
if err != nil {
    return err
}
rowsAffected, _ := result.RowsAffected()
log.Printf("Inserted %d rows", rowsAffected)

// Existing Flush() method continues to work as before
err := batch.Flush(ctx)
```

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -17,10 +17,11 @@ const DefaultBatchLimit = 1024
 // Batch collects the Added rows and executes in batches, after collecting Limit number of rows.
 // The default Limit is DefaultBatchLimit.
 type Batch struct {
-	Stmt        *sql.Stmt
-	values      []interface{}
-	rValues     []reflect.Value
-	size, Limit int
+	Stmt         *sql.Stmt
+	values       []interface{}
+	rValues      []reflect.Value
+	size, Limit  int
+	rowsAffected int64
 }
 
 // Add the values. The first call initializes the storage,
@@ -83,10 +84,13 @@ func (b *Batch) Add(ctx context.Context, values ...interface{}) error {
 // Size returns the buffered (unflushed) number of records.
 func (b *Batch) Size() int { return b.size }
 
-// FlushWithResult executes the statement, clears the storage, and returns the result.
-func (b *Batch) FlushWithResult(ctx context.Context) (sql.Result, error) {
+// RowsAffected returns the accumulated number of rows affected by all Flush operations.
+func (b *Batch) RowsAffected() int64 { return b.rowsAffected }
+
+// Flush executes the statement and clears the storage.
+func (b *Batch) Flush(ctx context.Context) error {
 	if len(b.rValues) == 0 || b.rValues[0].Len() == 0 {
-		return nil, nil
+		return nil
 	}
 
 	if b.values == nil {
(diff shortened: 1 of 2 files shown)


tgulacsi:
Please spare me these AI generated summaries!

This whole thing has a big problem: only FlushWithResult returns the RowsAffected,
Batch.Insert does not - though it will call Flush when the number of rows reaches the threshold.

So your FlushWithResult will return only the last parts' Result, which can be anything between 0 and Batch.Limit rows.

If you'd really want a propert RowsAffected, then you have to accumulate it in Batch on each FlushWithResult.

But then, should you always return the aggregated result, or only the current Flush's ?

Maybe the API would be more clear if Batch accumulated the RowsAffected, which would be queriable with a Result() method.

lukius:
Hi Tamás! Thanks for your feedback. Answers inlined below:

> Please spare me these AI generated summaries!

Although it was indeed AI-driven, I supervised and edited it before submitting. I think it conveys the main contributions in a clear way, but I completely understand if it falls outside your expectations.

> This whole thing has a big problem: only FlushWithResult returns the RowsAffected, Batch.Insert does not - though it will call Flush when the number of rows reaches the threshold.

Yes, it is true that `Add` still calls `Flush` and in those cases we miss partial results. As a matter of fact, this derived from how we use this feature in our services --we `Add` values never exceeding the default batch limit, and once we are done we just `Flush` everything to execute our prepared statement. Since we needed to check the affected row count, we reimplemented this batch struct locally. We then thought it would be a good contribution for the project.

> Maybe the API would be more clear if Batch accumulated the RowsAffected, which would be queriable with a Result() method.

That sounds good. I can revisit the code and get back to you once it is done.

tgulacsi (approved):

PR #402, opened 2026-02-26
Fix several small issues from code review

krisrice:
Critical fixes:
- subscr.go: Fix race condition in NewSubscription where subscriptionsID was read without the lock after unlock; capture localID before Unlock() to avoid the window where another goroutine could increment the counter.
- subscr.go: Fix C memory leaks — subscrID (C.malloc'd callback context) was never freed on dpiConn_subscribe error path or in Subscription.Close(). Added subscrID field to Subscription struct; free on both error and close. Also clean up subscriptions map on error path.
- lob.go: Guard all &p[0] dereferences against empty slices at five sites: dpiLobReader.ReadAt, dpiLobWriter.Write, DirectLob.Set, DirectLob.ReadAt, DirectLob.WriteAt. DirectLob.Set uses a nil *C.char for empty input to correctly clear the LOB rather than panicking.

High severity fixes:
- dsn/dsn.go: Fix typo poolMasSessionsPerShard -> poolMaxSessionsPerShard in both serialization (q.Add) and deserialization (parser table). The documented DSN parameter was silently ignored due to the missing 'x'.
- rows.go: Restore statement handle cleanup on getNumQueryColumns error path in both Next() and NextResultSet(). Replace commented-out C.dpiStmt_release calls with st.Close(), consistent with the openRows error path.
- obj.go: Replace panic() with error returns in three public API methods: GetAttribute (nil Object), GetItem (nil data), and ObjectType.init (nil drv). A library must never crash the caller process on bad input.
- json.go: Replace unchecked type assertions val.(JSONObject) and val.(JSONArray) with safe comma-ok assertions that return a descriptive error instead of panicking on unexpected types from d.Get().

Medium severity fixes:
- dsn/dsn.go: Document that sessionIncrement is a backwards-compat alias for poolIncrement; if both keys are present, sessionIncrement wins (order-dependent).
- drv.go, queue.go: Replace fmt.Printf in resource-leak finalizers with getLogger(context.Background()), falling back to fmt.Printf only when no logger is configured. Leak warnings now respect the configured log handler.
- conn.go: Replace discarded ok bool in atomic.Value type assertion (c.currentTT.Load().(TraceTag)) with explicit comma-ok check. Zero-value TraceTag remains the correct fallback but intent is now explicit.
- conn.go, data.go, drv.go, json.go, lob.go, obj.go, orahlp.go, queue.go, rows.go, subscr.go, token.go: Replace all 40+ context.TODO() occurrences with context.Background(). These are all logger-only call sites with no real context available; Background() is the correct semantic ("intentionally no deadline") vs TODO() ("known placeholder"). log.go is intentionally unchanged as it contains the ctx != context.TODO() special-case guard.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -149,7 +149,7 @@ func (c *conn) Break() error {
 	}
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	logger := getLogger(context.TODO())
+	logger := getLogger(context.Background())
 	if logger != nil {
 		logger.Debug("Break", "dpiConn", c.dpiConn)
 	}
@@ -433,7 +433,7 @@ func (c *conn) newVar(vi varInfo) (*C.dpiVar, []C.dpiData, error) {
 	}
 	var dataArr *C.dpiData
 	var v *C.dpiVar
-	logger := getLogger(context.TODO())
+	logger := getLogger(context.Background())
 	if logger != nil {
 		logger.Debug("dpiConn_newVar", "conn", c.dpiConn, "typ", int(vi.Typ), "natTyp", int(vi.NatTyp), "sliceLen", vi.SliceLen, "bufSize", vi.BufSize, "isArray", isArray, "objType", vi.ObjectType, "v", v)
 	}
@@ -496,7 +496,7 @@ func (c *conn) init(ctx context.Context, isNew bool, onInit func(ctx context.Con
 }
 
 func (c *conn) initTZ() error {
-	logger := getLogger(context.TODO())
+	logger := getLogger(context.Background())
 	if logger != nil {
 		logger.Debug("initTZ", "tzValid", c.tzValid, "paramsTZ", c.params.Timezone)
 	}
@@ -626,7 +626,7 @@ func maybeBadConn(err error, c *conn) error {
 		return nil
 	}
 	cl := func() {}
-	ctx := context.TODO()
+	ctx := context.Background()
 	logger := getLogger(ctx)
 	if c != nil {
 		cl = func() {
@@ -717,7 +717,10 @@ func (c *conn) setTraceTag(tt TraceTag) error {
 		return nil
 	}
 	todo := make([][2]string, 0, 5)
-	currentTT, _ := c.currentTT.Load().(TraceTag)
+	var currentTT TraceTag
+	if v, ok := c.currentTT.Load().(TraceTag); ok {
+		currentTT = v
+	}
 	for nm, vv := range map[string][2]string{
 		"action":     {currentTT.Action, tt.Action},
 		"module":     {currentTT.Module, tt.Module},
@@ -973,7 +976,7 @@ func (c *conn) IsValid() bool {
 	if dpiConnOK {
 		dpiConnOK = c.isHealthy()
 	}
-	logger := getLogger(context.TODO())
+	logger := getLogger(context.Background())
 	if logger != nil {
 		logger.Debug("IsValid", "connOK", dpiConnOK, "released", released, "pooled", pooled, "tzOK", tzOK)
 	}
--- a/json.go
+++ b/json.go
@@ -341,7 +341,7 @@ func (j JSON) GetValue(opts JSONOption) (any, error) {
 	} else {
 		return val, nil
 	}
-	if logger := getLogger(context.TODO()); logger != nil {
+	if logger := getLogger(context.Background()); logger != nil {
 		logger.Error("JSON.GetValue", "error", err)
 	}
 	return nil, err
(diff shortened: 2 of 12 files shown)


lob.go
```diff
@@ -534,8 +540,12 @@ func (dl *DirectLob) Trim(size int64) error {
 // Set the contents of the LOB to the given byte slice.
 // The LOB is cleared first.
 func (dl *DirectLob) Set(p []byte) error {
+	var ptr *C.char
+	if len(p) > 0 {
```
tgulacsi: len(p) != 0

obj.go
```diff
@@ -99,7 +99,7 @@ func (O *Object) SetAttribute(name string, data *Data) error {
 		}
 		// name = try
 	}
-	// ctx := context.TODO()
+	// ctx := context.Background()
 	// logger := getLogger(ctx)
 	// if logger != nil {
 	// 	logger = logger.With("object", O.Name, "name", name)
```
tgulacsi: Just delete the commented out lines.

tgulacsi (changes requested):
If you could split this to smaller changes like:
- context.TODO -> Background
- subscrID
- fmt -> logger
- poolMas
- rest

then I could approve the first three as it arrives, and discuss the rest later...

conn.go
```diff
@@ -717,7 +717,10 @@ func (c *conn) setTraceTag(tt TraceTag) error {
 		return nil
 	}
 	todo := make([][2]string, 0, 5)
-	currentTT, _ := c.currentTT.Load().(TraceTag)
```
tgulacsi, line 720: NAK, this is inherent.

json.go
```diff
@@ -432,11 +432,19 @@ func (j JSONScalar) GetValue() (val any, err error) {
 	} else {
 		val = d.Get()
 		if j.dpiJsonNode.oracleTypeNum == C.DPI_ORACLE_TYPE_JSON_OBJECT {
-			jobj := val.(JSONObject)
```
tgulacsi, line 435: NAK, this is inherent (the type must be JSONObject).

json.go
```diff
@@ -432,11 +432,19 @@ func (j JSONScalar) GetValue() (val any, err error) {
...
+			jobj, ok := val.(JSONObject)
+			if !ok {
+				err = fmt.Errorf("expected JSONObject, got %T", val)
+				return
+			}
 			jobj.stringOption = j.stringOption
 			val, err = jobj.GetValue()
 		} else if j.dpiJsonNode.oracleTypeNum == C.DPI_ORACLE_TYPE_JSON_ARRAY {
-			jarr := val.(JSONArray)
```
tgulacsi, line 439: NAK, this is inherent (the type must be JSONArray).

lob.go
```diff
@@ -378,6 +378,9 @@ func (dlr *dpiLobReader) ReadAt(p []byte, off int64) (int, error) {
 	if dlr.IsClob {
 		return 0, ErrCLOB
 	}
+	if len(p) == 0 {
```
tgulacsi, line 381: This should be a panic (programmer error).

lob.go
```diff
@@ -552,6 +562,9 @@ func (dl *DirectLob) Set(p []byte) error {
 // For this reason, if a character requires more than one UCS-2 codepoint,
 // the size returned will be inaccurate and care must be taken to account for the difference!
 func (dl *DirectLob) ReadAt(p []byte, offset int64) (int, error) {
+	if len(p) == 0 {
```
tgulacsi, line 565: panic, programmer error.

obj.go
```diff
@@ -51,7 +51,7 @@ var ErrNoSuchKey = errors.New("no such key")
 // GetAttribute gets the i-th attribute into data.
 func (O *Object) GetAttribute(data *Data, name string) error {
 	if O == nil {
-		panic("nil Object")
+		return errors.New("nil Object")
```
tgulacsi, line 54: NAK, programmer error.

obj.go
```diff
@@ -886,7 +886,7 @@ func (O ObjectCollection) GetItem(data *Data, i int) error {
 		return ErrNotCollection
 	}
 	if data == nil {
```
tgulacsi, line 888: NAK, programmer error.

rows.go
```diff
@@ -828,13 +828,13 @@ func (r *rows) NextResultSet() error {
...
 	var n C.uint32_t
-	logger := getLogger(context.TODO())
+	logger := getLogger(context.Background())
 	if err := r.checkExec(func() C.int { return C.dpiStmt_getNumQueryColumns(st.dpiStmt, &n) }); err != nil {
 		err = fmt.Errorf("getNumQueryColumns: %+v: %w", err, io.EOF)
 		if logger != nil {
 			logger.Error("NextResultSet.getNumQueryColumns", "st", fmt.Sprintf("%p", st.dpiStmt), "error", err)
 		}
-		//C.dpiStmt_release(st.dpiStmt)
+		st.Close()
```
tgulacsi, line 837: NAK, this will lock.

tgulacsi (changes requested):

PR #408, opened 2026-05-20
Return  the Oracle Query ID on subscription registration

ddelpero:
subscr.go: return the Oracle Query ID on subscription registration

Returned the Oracle CQN query ID by updating the `Register` method signature to return both the `uint64` ID and an `error`. 

Previously, the dynamic query ID assigned by the Oracle database kernel was not returned, making it difficult to dynamically route incoming `godror.Event` updates when managing multiple granular queries on a single physical table.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,3 +23,5 @@ require (
 )
 
 // retract v0.45.0
+
+replace github.com/godror/godror => github.com/ddelpero/godror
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -23,3 +23,5 @@ require (
 )
 
 // retract v0.45.0
+
+replace github.com/godror/godror => github.com/ddelpero/godror
```
tgulacsi, line 27: Revert this change.

subscr.go
```diff
@@ -275,7 +275,7 @@ func (c *conn) NewSubscription(name string, cb func(Event), options ...Subscript
 // Register a query for Change Notification.
 //
 // This code is EXPERIMENTAL yet!
-func (s *Subscription) Register(qry string, params ...any) error {
+func (s *Subscription) Register(qry string, params ...any) (uint64, error) {
```
tgulacsi, line 278: Document what that returned uint64 is, and how can it be used.

tgulacsi (changes requested):
BTW squash all the intermediate commits!

PR #416, opened 2026-08-11
fix(conn): don't leak pooled session on bad-conn close; keep session in IsValid

### LINKED ISSUE #415: Session pool exhaustion (ORA-24496) never recovers after context-cancelled queries
**Describe the bug**

Pooled sessions leak permanently — `dpiPool` `busy` count never decreases — when `database/sql` force-closes a connection while a db-bound prepared statement (`db.PrepareContext`) still has a cached driver statement handle on it. Each occurrence pins one session as "busy" forever. Once all `MaxSessions` are pinned, every subsequent acquire fails with `ORA-24496: OCISessionGet() timed out waiting for a free connection` and the process never recovers (`maxLifeTime` / `sessionTimeout` do not help: they only apply to sessions that are actually released). The only remedy is a process restart.

The trigger is ordinary production traffic: HTTP clients cancelling requests mid-query (context cancellation) under concurrency.

Root-cause chain (verified with `DPI_DEBUG_LEVEL=2` refcount traces and a `SIGQUIT` goroutine dump showing the process completely idle while `busy=10`):

1. `conn.IsValid()` (called by `database/sql` before pooling an idle connection) releases the underlying session for pooled connections (`closeNotLocking` → `dpiConn_release`, the issue #57 mitigation). From that point the session is kept alive *only* by child handle references — e.g. a `dpiStmt` belonging to a `*sql.Stmt` prepared through `db.PrepareContext`, whose driver statements `database/sql` caches per connection.
2. When a query on such a connection fails hard (context cancellation → `Break` → error mapped to `driver.ErrBadConn`), `database/sql` force-closes the connection. Its `(*Stmt).Close` → `finalClose` → `noteUnusedDriverStatement` then **skips** closing the cached driver statement because the connection is already `finalClosed` (see `database/sql/sql.go`), assuming the connection close freed driver-side resources.
3. godror's `conn.Close`/`closeNotLocking` is only a `dpiConn_release` (refcount decrement). The orphaned `dpiStmt` keeps the `dpiConn` refcount above zero, so the session is never returned to the pool: `busy` stays incremented forever.

A closely-related second defect makes things worse: `conn.ResetSession` claims ("ResetSession will reacquire it") but never re-acquires a session for a connection that `IsValid` released — it just returns `driver.ErrBadConn` because `c.dpiConn == nil`. So pooled connections are never actually reused, and in the `database/sql` direct-handoff path (pool at `MaxOpenConns` with waiters queued) that `ErrBadConn` reaches application code as `driver: bad connection` (we measured 8–17% request failures at 40 concurrent users, with zero cancellations involved).

**To Reproduce**

Minimal test case below. It opens a pooled godror `sql.DB` (`MaxSessions=10`), then runs waves of 20 concurrent `db.PrepareContext` + `stmt.QueryContext` calls whose contexts are cancelled 10–90 ms in (simulating clients disconnecting), probing pool health between waves:

```
DB_USERNAME=scott \
DB_PASSWORD=tiger \
DB_DSN='(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=db.example.com)(PORT=1521))(CONNECT_DATA=(SERVER=POOLED)(SERVICE_NAME=orclpdb1.example.com)))' \
go run .
```

```go
package main

import (
        "context"
        "database/sql"
        "fmt"
        "math/rand"
        "os"
        "strings"
        "sync"
        "time"

        "github.com/godror/godror"
)

const query = `SELECT object_name FROM all_objects WHERE rownum <= :1`

func main() {
        cs, err := godror.ParseDSN(os.Getenv("DB_DSN"))
        check(err)
        cs.Username = os.Getenv("DB_USERNAME")
        cs.Password = godror.NewPassword(os.Getenv("DB_PASSWORD"))
        cs.StandaloneConnection = godror.Bool(false)
        cs.PoolParams = godror.PoolParams{
                MinSessions:    3,
                MaxSessions:    10,
                WaitTimeout:    3 * time.Second,
                MaxLifeTime:    15 * time.Minute,
                SessionTimeout: 10 * time.Minute,
        }

        db, err := sql.Open("godror", cs.StringWithPassword())
        check(err)
        db.SetMaxIdleConns(3)
        db.SetMaxOpenConns(10)

        check(probe(db))
        fmt.Println("initial probe OK")

        for wave := 1; wave <= 30; wave++ {
                var wg sync.WaitGroup
                for i := 0; i < 20; i++ {
                        wg.Add(1)
                        go func() {
                                defer wg.Done()
                                // Cancel mid-query, like an HTTP client disconnecting.
                                ctx, cancel := context.WithTimeout(context.Background(),
                                        time.Duration(10+rand.Intn(80))*time.Millisecond)
                                defer cancel()

                                // db-bound prepared statement (db.PrepareContext), as any
                                // PrepareContext+QueryContext helper does.
                                stmt, err := db.PrepareContext(ctx, query)
                                if err != nil {
                                        return
                                }
                                defer stmt.Close()

                                rows, err := stmt.QueryContext(ctx, 50000)
                                if err != nil {
                                        return
                                }
                                defer rows.Close()

                                var s string
                                for rows.Next() {
                                        _ = rows.Scan(&s)
                                }
                        }()
                }
                wg.Wait()
                time.Sleep(500 * time.Millisecond)

                if err := probe(db); err != nil {
                        fmt.Printf("wave %d: probe FAILED: %v\n", wave, err)
                        if strings.Contains(err.Error(), "ORA-24496") {
                                fmt.Println("\npool exhausted; sleeping 60s with zero load to show it never recovers...")
                                time.Sleep(60 * time.Second)
                                fmt.Printf("after 60s idle, probe: %v\n", probe(db))
                                os.Exit(1)
                        }
                } else {
                        fmt.Printf("wave %d: probe OK\n", wave)
                }
        }
        fmt.Println("no leak detected after 30 waves")
}

// probe runs one trivial query with a generous deadline; when the pool is
// fully pinned it fails with ORA-24496 after the pool's waitTimeout.
func probe(db *sql.DB) error {
        ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
        defer cancel()
        var one int
        return db.QueryRowContext(ctx, "SELECT 1 FROM DUAL").Scan(&one)
}

func check(err error) {
        if err != nil {
                panic(err)
        }
}
```

Output on our environment:

```
initial probe OK
wave 1: probe OK
wave 2: probe OK
wave 3: probe OK
wave 4: probe FAILED: pool=0x3e885e50 stats=busy=10 open=10 max=10 maxLifetime=15m0s timeout=10m0s waitTimeout=3s params={...}: ORA-24496: OCISessionGet() timed out waiting for a free connection

pool exhausted; sleeping 60s with zero load to show it never recovers...
after 60s idle, probe: pool=0x3e885e50 stats=busy=10 open=10 max=10 ...: ORA-24496: OCISessionGet() timed out waiting for a free connection
```

Reproduced identically with `(SERVER=POOLED)` (DRCP) and `(SERVER=DEDICATED)` in the DSN — the leak is client-side, independent of server-side pooling. Also reproduced on godror v0.49.6, v0.50.0 and v0.51.0, and with Instant Client 23ai (el9) and the 26ai release repo (el10).

**Expected behavior**

Sessions always return to the pool once no query is running. A connection close should free its session even if `database/sql` abandoned a cached statement handle on it (or `IsValid`/`ResetSession` should keep the session attached to the pooled connection so that cached statements stay valid and connections are genuinely reusable).

We validated a two-line direction that fixes both defects for us (happy to send a PR):

1. `IsValid()`: keep the session attached for pooled connections (don't release) — `ResetSession` then finds a healthy `dpiConn` and connections are actually reused; cached prepared statements stay valid.
2. `closeNotLocking()`: call `C.dpiConn_close(dpiConn, C.DPI_MODE_CONN_CLOSE_DEFAULT, nil, 0)` before `C.dpiConn_release(dpiConn)`, so a genuinely-closed connection returns its session even when orphaned child handles keep the refcount above zero.

With that patch the same workload runs at 121k requests / 100.00% success / 674 req/s, survives repeated cancellation storms, and the pool always recovers.

**Error output**

The ORA-24496 pool dump is shown in the reproducer output above (it repeats every `waitTimeout` forever once bricked).

`DPI_DEBUG_LEVEL=2` refcount trace of one leaked statement (created → rows opened (addRef → 2) → rows closed (release → 2→1) → **final release never happens**; exactly `MaxSessions` such `dpiStmt` handles exist at refcount 1 once the pool is bricked, each pinning one `dpiConn`):

```
ODPI [00019] 2026-08-10 15:55:53.420: fn start dpiStmt_addRef(0x7fb2b4032980)
ODPI [00019] 2026-08-10 15:55:53.420: ref 0x7fb2b4032980 (dpiStmt) -> 2
ODPI [00019] 2026-08-10 15:55:53.420: fn end dpiStmt_addRef(0x7fb2b4032980) -> 0
ODPI [00019] 2026-08-10 15:55:53.420: fn start dpiStmt_fetchRows(0x7fb2b4032980)
ODPI [00019] 2026-08-10 15:55:53.420: fn end dpiStmt_fetchRows(0x7fb2b4032980) -> 0
ODPI [00019] 2026-08-10 15:55:53.421: fn start dpiStmt_release(0x7fb2b4032980)
ODPI [00019] 2026-08-10 15:55:53.421: ref 0x7fb2b4032980 (dpiStmt) -> 1
ODPI [00019] 2026-08-10 15:55:53.421: fn end dpiStmt_release(0x7fb2b4032980) -> 0
(no further lines for 0x7fb2b4032980 — statement.Close is never called)
```

A `SIGQUIT` goroutine dump taken while `busy=10` shows the process fully idle: only runtime, `database/sql` `connectionOpener`/`connectionCleaner`, and HTTP server goroutines — nothing in godror or cgo. The sessions are held by leaked handles, not by running calls.

**Your oracle client version**

oracle-instantclient-basiclite 23.26.3.0.0 (installed from the `oracle-instantclient-release-26ai-el10` repo; ODPI-C 6.0.0). Also reproduced with instantclient-basiclite 23ai on el9.

Database version: Oracle AI Database 26ai Enterprise Edition Release 23.26.2.0.0 - Production

**Your godror version**

v0.51.0 (also reproduced with v0.50.0 and v0.49.6)

**Your go version**

go1.26.5 (Red Hat 1.26.5-1.el10_2) linux/amd64

**Your gcc version**

gcc (GCC) 14.3.1 20251022 (Red Hat 14.3.1-4)

**Machine (please complete the following information):**

 - OS: Linux (Oracle Linux Server 10.2 container; build in ubi10/go-toolset)
 - Architecture: x86_64
 - Version: kernel 7.0.14 (Fedora 44 host running the containers)

**Additional context**

- Not configuration-dependent: with `StandaloneConnection=false` the reproducer bricks the pool regardless of `database/sql` settings (`SetMaxIdleConns` 0, 3 or 10 — all brick in 3–4 waves). With `StandaloneConnection=true` (no client-side session pool) the same workload survives 30 waves (~600 cancelled queries) with no leak, which points at the pooled-session release path rather than application usage.
- `SERVER=POOLED` (DRCP) vs `SERVER=DEDICATED` in the DSN makes no difference.
- The database is reached over a WAN in our setup, which makes mid-query cancellations (client disconnects) and `Break`-related connection failures more common, but the reproducer bricks the pool even on a fast link — it only needs a query slow enough (~50–200 ms) for the cancellation to land mid-flight.
- Related: issue #57 (the `IsValid` release was introduced as its mitigation). The mitigation trades that bug for this leak: it makes the session's survival depend on statement handle refcounts, and `database/sql` legitimately abandons statement handles on force-closed connections (`noteUnusedDriverStatement` skips `ds.Close()` when the connection is `finalClosed`).

amsal:
Fixes #415. 

Cancelled queries can pin a pooled session as busy forever -> ORA-24496 on every subsequent acquire, unrecoverable without a restart. Also surfaces as "driver: bad connection" under pool contention.

Test: TestPoolLeakOnCancel. StandaloneConnection=true is unaffected.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -633,6 +633,13 @@ func maybeBadConn(err error, c *conn) error {
 			if logger != nil {
 				logger.Error("maybeBadConn close", "conn", c, "error", err)
 			}
+			// dpiConn_close before release so an orphaned dpiStmt (abandoned by
+			// database/sql on a force-closed conn) can't pin the pooled session busy
+			// forever (ORA-24496, #415). Only the bad-conn path hard-closes; healthy
+			// Close keeps the plain release (queues/LOBs outlive the conn via refcount).
+			if c.dpiConn != nil {
+				C.dpiConn_close(c.dpiConn, C.DPI_MODE_CONN_CLOSE_DEFAULT, nil, 0)
+			}
 			_ = c.closeNotLocking()
 		}
 	}
@@ -961,8 +968,8 @@ func (c *conn) ResetSession(ctx context.Context) error {
 // If implemented, drivers may return the underlying error from queries,
 // even if the connection should be discarded by the connection pool.
 //
-// This implementation returns the underlying session to the OCI session pool,
-// iff this is a pooled connection. ResetSession will reacquire it.
+// A pooled connection keeps its session so it stays reusable; the session
+// returns to the OCI pool only when the connection is closed.
 func (c *conn) IsValid() bool {
 	if c == nil {
 		return false
@@ -988,16 +995,6 @@ func (c *conn) IsValid() bool {
 		return dpiConnOK
 	}
 
-	// FIXME(tgulacsi): Prepared statements hold the previous session,
-	// so sometimes sessions are not released, resulting in
-	//
-	//     ORA-24459: OCISessionGet()
-	//
-	// See https://github.com/godror/godror/issues/57 for example.
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	_ = c.closeNotLocking()
-	c.released = true
 	return true
 }
 
--- a/z_i415_test.go
+++ b/z_i415_test.go
@@ -1,0 +1,59 @@
+package godror_test
+
+import (
+	"context"
+	"database/sql"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/godror/godror"
+	"github.com/godror/godror/dsn"
+)
+
+// TestPoolLeakOnCancel reproduces the ORA-24496 pool leak: a query cancelled
+// mid-execute must not pin its pooled session.
+func TestPoolLeakOnCancel(t *testing.T) {
+	P, err := dsn.Parse(testConStr)
+	if err != nil {
+		t.Fatal(err)
+	}
+	P.StandaloneConnection = godror.Bool(false)
+	P.PoolParams.MinSessions, P.PoolParams.MaxSessions = 2, 4
+	P.PoolParams.WaitTimeout = 3 * time.Second
+
+	db := sql.OpenDB(godror.NewConnector(P))
+	defer db.Close()
+	db.SetMaxOpenConns(P.PoolParams.MaxSessions)
+
+	ctx, cancel := context.WithTimeout(testContext("PoolLeakOnCancel"), time.Minute)
+	defer cancel()
+
+	probe := func(ctx context.Context) error {
+		var n int
+		return db.QueryRowContext(ctx, "SELECT 1 FROM DUAL").Scan(&n)
+	}
+	if err := probe(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	var wg sync.WaitGroup
+	for i := 0; i < 200; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			cctx, ccancel := context.WithTimeout(ctx, time.Duration(1+i%20)*time.Millisecond)
+			defer ccancel()
+			_, _ = db.ExecContext(cctx, "BEGIN DBMS_SESSION.SLEEP(1); END;")
+		}(i)
+	}
+	wg.Wait()
+
+	pctx, pcancel := context.WithTimeout(ctx, 15*time.Second)
+	defer pcancel()
+	for i := 0; i < 3; i++ {
+		if err := probe(pctx); err != nil {
+			t.Fatalf("pool did not recover after cancelled queries: %+v", err)
+		}
+	}
+}


conn.go
```diff
@@ -252,6 +252,20 @@ func (c *conn) closeNotLocking() error {
...
 
+// closeHardNotLocking closes the conn before releasing it, so an orphaned
+// dpiStmt (abandoned by database/sql on a force-closed conn) can't pin the
+// pooled session busy forever (ORA-24496). maybeBadConn only; healthy Close
+// keeps the plain release (queues/LOBs outlive the conn via refcount).
+func (c *conn) closeHardNotLocking() error {
+	if c == nil {
+		return nil
+	}
+	if dpiConn := c.dpiConn; dpiConn != nil {
```
tgulacsi: why not just 
if c.dpiConn != nil {
...

tgulacsi: Even better: as you use this new closeHardNotLocing function only at one place, "inline" it there: just call C.dpiConn_close before allign c.closeNotLocking in maybeBadConn.
The comment fits there too.

z_i415_test.go
```diff
@@ -0,0 +1,71 @@
...
+		t.Fatal(err)
+	}
+
+	const qry = "SELECT object_name FROM all_objects WHERE rownum <= :1"
+	var wg sync.WaitGroup
+	for i := 0; i < 200; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			cctx, ccancel := context.WithTimeout(ctx, time.Duration(1+i%20)*time.Millisecond)
```
tgulacsi, line 45: Is this reliably enough for "SELECT 1 FROM DUAL"?
If not, a "BEGIN DBMS_SESSION.SLEEP(1); END;" could be more fruitful.

tgulacsi (changes requested):

amsal:
Updated with changes

tgulacsi (approved):

PR #417, opened 2026-08-20
Allow token callbacks during connection creation

nihar-oracle:
## Summary
- Keep `dpiConn_create` and `dpiPool_create` callback-capable because `TokenCB` may invoke Go from ODPI-C during connection or pool creation.
- Leave the two `nocallback` directives visibly commented beside that rationale, rather than adding a source-scanning regression test.

## Validation
- **Passed:** package compilation with the Go 1.24+ cgo preamble and no test execution.
  - Command: `DO_NOT_CONNECT=true go test -run '^$' .`
- **Environment-limited:** `DO_NOT_CONNECT=true go test .` could not complete because the suite still exercises Oracle-dependent tests and no local Oracle Client/DPI context is available.


nihar-oracle:
@tgulacsi @sudarshan12s, could you please review this change? It touches the Go 1.24 cgo annotations and TokenCB paths you respectively introduced. Thank you.

tgulacsi (changes requested):
This test concentrates on one file, which will be renamed/removed as go 1.24 becames old (now, as 1.27 has been released).
Either parse all non *_test.go file, or just skip this and comment the reason of why those functions should not be nocallback.
I vote for the second....

tgulacsi (changes requested):
Comment it out (remove #or add a space after it) with a comment about the reason.

nihar-oracle:
Addressed the requested changes: removed `drv_go24_test.go` and retained the `dpiConn_create` and `dpiPool_create` directives as commented lines with adjacent TokenCB callback rationale. Validation passed: `DO_NOT_CONNECT=true go test -run '^$' .`.

drv_go24.go
```diff
@@ -13,7 +13,8 @@ package godror
 /*
 #cgo nocallback dpiConn_breakExecution
 #cgo nocallback dpiConn_commit
-#cgo nocallback dpiConn_create
+// TokenCB may invoke Go from ODPI-C during connection creation, so permit callbacks.
```
sudarshan12s: may be with more specific: 

When acquiring a pooled connection, it can invoke the access-token callback implemented in Go (`TokenCB`).

drv_go24.go
```diff
@@ -159,7 +160,8 @@ package godror
 #cgo nocallback dpiObjectType_getInfo
 #cgo nocallback dpiObjectType_release
 #cgo nocallback dpiPool_close
-#cgo nocallback dpiPool_create
+// TokenCB may invoke Go from ODPI-C during pool creation, so permit callbacks.
```
sudarshan12s: It installs the access-token callback before creating the session pool, and OCI may invoke the callback implemented in Go (`TokenCB`) during pool creation.

sudarshan12s (commented):
added minor changes to comments to make it precise.

drv_go24.go
nihar-oracle: Thank you—made this comment precise to state that acquiring a pooled connection through dpiConn_create can invoke the Go TokenCB.

drv_go24.go
nihar-oracle: Thank you—made this comment precise to state that dpiPool_create installs the callback before pool creation and OCI may invoke the Go TokenCB during that creation.

nihar-oracle:
> This test concentrates on one file, which will be renamed/removed as Go 1.24 becomes old. Either parse all non-`*_test.go` files, or skip this and comment why those functions should not be `nocallback`. I vote for the second.

@tgulacsi Thank you for the review. We followed your second option: removed `drv_go24_test.go` rather than parsing non-test source files, and kept only the precise adjacent rationale comments on the two commented-out directives.

tgulacsi:
Have you tested this last version?
I'm not sure about what kind of #cgo incantations are picked up and which not - to be sure (if cannot test), I'd obfuscate it totally ( like "# c g o").

nihar-oracle:
> Have you tested this last version?
> I'm not sure about what kind of #cgo incantations are picked up and which not - to be sure (if cannot test), I'd obfuscate it totally ( like "# c g o").

@tgulacsi Yes. I tested the exact preamble form with Go 1.26.6: a minimal callback program using `// #cgo nocallback callGo` returned normally, while the active `#cgo nocallback callGo` control panicked with `function marked with #cgo nocallback called back into Go`. The branch also passes `DO_NOT_CONNECT=true go test -run '^$' .`.

I still applied the unambiguous spelling you suggested to both disabled directives (`# c g o`) and pushed it in `0f74148`.

tgulacsi (approved):

Issue #400, opened 2026-01-26
malloc(): unsorted double linked list corrupted in CGO


p4r53c:
**Describe the bug**
We encountered strange behavior in `v0.49.3`. We started reporting cases where, strictly in a container environment (Docker), our application panic in `CGO` layer calls to the `dpiConn_create()` C function in ODPI-C due to memory corruption on allocations.

**Error output**
 ```
malloc(): unsorted double linked list corrupted
 SIGABRT: abort
 PC=0x7f305ea8902c m=7 sigcode=18446744073709551610
 signal arrived during cgo execution
 
 goroutine 467607 gp=0xc000773c00 m=7 mp=0xc000580008 [syscall, locked to thread]:
 runtime.cgocall(0x2303930, 0xc000ccafc0)
 	/var/jenkins_home/tools/org.jenkinsci.plugins.golang.GolangInstallation/go-1.25.4/src/runtime/cgocall.go:167 +0x4b fp=0xc000ccaf98 sp=0xc000ccaf60 pc=0x49712b
 github.com/godror/godror._Cfunc_dpiConn_create(0x7f3008081e70, 0x7f2ff815ae10, 0x8, 0x7f2ff80ddb30, 0x14, 0x7f2ff82917f0, 0x33, 0xc000b554f0, 0xc000aef0e0, 0xc000a94ca0)
 	/tmp/go-build4220533701/b689/_cgo_gotypes.go:2881 +0x5d fp=0xc000ccafc0 sp=0xc000ccaf98 pc=0x1e9c49d
 github.com/godror/godror.(*drv).acquireConn.func8.1(...)
 	goroot/pkg/mod/github.com/godror/godror@v0.49.3/drv.go:706
 github.com/godror/godror.(*drv).acquireConn.func8()
 	goroot/pkg/mod/github.com/godror/godror@v0.49.3/drv.go:706 +0x9c5 fp=0xc000ccb360 sp=0xc000ccafc0 pc=0x1eb5ca5
 github.com/godror/godror.(*drv).checkExecNoLOT(0x4ce9ac0, 0x7f2ff8291701?)
 	goroot/pkg/mod/github.com/godror/godror@v0.49.3/drv.go:286 +0x1b fp=0xc000ccb378 sp=0xc000ccb360 pc=0x1eb1f1b
 github.com/godror/godror.(*drv).checkExec(0x4ce9ac0, 0xc000ccb658)
[...]

Issue #361, opened 2025-02-28
double free or corruption when running queries in parallel


anders-swanson:
**Describe the bug**
When running multiple queries in parallel, godror sometimes hits a SIGABRT double free or corruption error  attempting to create a connection during cgo execution with the `godror._Cfunc_dpiConn_create` function.

Note: I've only been noticing this on Oracle Linux 8. Doesn't seem to matter if it's amd64 or arm architecture.

**To Reproduce**
I reproduced this error in a sample repo: [https://github.com/anders-swanson/godror-sample](https://github.com/anders-swanson/godror-sample). 

After you download the sample, run it like so (requires docker compatible container runtime, docker-compose):

```
docker build . -t container-registry.oracle.com/database/godror-sample:latest && docker-compose up
```

I see the error every time with the default settings in the `compose.yaml` file. 

**Expected behavior**
cgo errors of this type either don't occur or are handled better by godror.

**Error output**

```
sample      | double free or corruption (out)
sample      | SIGABRT: abort
sample      | PC=0x7f249fcad5ef m=0 sigcode=18446744073709551610
sample      | signal arrived during cgo execution
sample      |
sample      | goroutine 1574 gp=0xc00031bc00 m=0 mp=0xaafec0 [syscall, locked to thread]:
sample      | runtime.cgocall(0x644580, 0xc0004940c8)
sample      | 	/usr/local/go/src/runtime/cgocall.go:157 +0x4b fp=0xc0004940a0 sp=0xc000494068 pc=0x4135eb
sample      | github.com/godror/godror._Cfunc_dpiConn_create(0x2b30000, 0x2bb54a0, 0x8, 0x2b9a360, 0xc, 0x2b705a0, 0x18, 0xc00040caf0, 0xc000419d40, 0xc000412980)
sample      | 	_cgo_gotypes.go:2672 +0x4b fp=0xc0004940c8 sp=0xc0004940a0 pc=0x5c2f6b
sample      | github.com/godror/godror.(*drv).acquireConn.func8.1(0xc000494198?, 0x5c97c8?, {0xc00001e032?, 0x8}, 0x7f24a0640108?, {0xc00001e012?, 0xc}, 0xc000412980?, 0x0?, 0xc00040caf0, ...)
sample      | 	/root/go/pkg/mod/github.com/godror/godror@v0.46.0/drv.go:680 +0x146 fp=0xc000494158 sp=0xc0004940c8 pc=0x5d6606
[...]

PR #418, opened 2026-08-27
Serialize standalone connection creation

CBEPX:
Fixes the `malloc(): unsorted double linked list corrupted` / `double free or corruption` -> `SIGABRT` inside `dpiConn_create` reported in #361 and #400.

**What happens.** Oracle Client resolves the OS user name inside `OCIServerAttach` (`osncon -> niqname -> niqnamcd -> snigun`). When `getpwuid()` fails for the process uid -- that is, the uid has no `/etc/passwd` entry, which is the normal state of a container running as a numeric `USER` -- and several goroutines attach at the same time, the client double-frees a process-shared buffer. The next allocation aborts the process inside `dpiConn_create`. Standalone connections expose this because every connection attaches on its own OCI environment; pooled connections attach under the pool and are not affected. This matches the "only reproducible inside a container" observation in #361.

**Evidence** (linux/amd64, glibc 2.36, Go 1.26.1, godror v0.51.4, Instant Client 21.12; the server side is a TNS mock, so no real database is required):

- `go build -asan`: `AddressSanitizer: attempting double-free` on an 8192-byte region, the freeing stack entirely inside `libclntsh.so`: `OCIServerAttach -> kpuatch -> upiini -> kpuadef -> osncon -> niqname -> niqnamcd -> snigun -> free()`, called from `dpiOci__serverAttach` (dpiOci.c:3171) <- `dpiConn__createStandalone` (dpiConn.c:474).
- Same binary, same uid without a passwd entry: abort within 1-3 s on every run; after appending a passwd line for that uid: 187 rounds, 5977 connects, 0 errors.
- With this patch, on the same uid without a passwd entry: `TestStandaloneChurn` passes (187 rounds / 5977 connects / 0 errors); unpatched v0.51.4 fails the same test in ~1.3 s.

**The change.** A package-level mutex serialises `dpiConn_create` for standalone connections only (`pool == nil`). Release stays unlocked: `dpiConn_release` performs no user-name lookup and only drops a refcount, and locking it would stall every standalone `Close` behind an in-flight connect for the client's connect timeout. Pooled connections are untouched.

The real fix for affected deployments is a passwd entry for the runtime uid (or a session pool); the mutex keeps the driver from triggering the client bug for everyone else. Happy to put it behind a build tag or a dsn option if you prefer.

**Test.** `z_standalone_churn_test.go` adds an opt-in subprocess churn test (`GODROR_TEST_STANDALONE_CHURN=1`, ~1 minute, mirrors `z_conc_test.go`); a native abort in the child fails the test instead of killing the runner. The faulty client path is only exercised when the process uid has no passwd entry; the test logs a notice otherwise.

### DIFF
--- a/drv.go
+++ b/drv.go
@@ -743,6 +743,24 @@ func (d *drv) createConn(pool *connPool, P commonAndConnParams) (*conn, bool, er
 	return &c, isNew, nil
 }
 
+// standaloneAttachMu serialises the creation of standalone connections.
+// Oracle Client resolves the OS user name inside OCIServerAttach
+// (osncon/niqname/snigun); when getpwuid() fails for the process uid (no
+// /etc/passwd entry, common in containers running as a numeric USER) and
+// several threads attach at the same time, it double-frees a process-shared
+// buffer and the process aborts inside dpiConn_create with glibc allocator
+// errors (#361, #400, oracle-db-appdev-monitoring#384; confirmed with ASan).
+// Pooled connections attach under the pool and are not affected. Release is
+// not locked: dpiConn_release never runs the user-name lookup, and it only
+// drops a refcount; the env is freed by whichever child handle releases
+// last. The real fix is a passwd entry for the uid; this lock keeps
+// standalone connections from triggering the client bug.
+//
+// The lock is process-wide because the buffer is: one slow or unreachable
+// standalone connect stalls every other standalone connect in the process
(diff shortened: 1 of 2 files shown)


tgulacsi (approved):

tgulacsi:
A bit verbose for my taste, but OK.

PR #422, opened 2026-09-19
test: add regression test for #421 charset commonCreateParams leak

### LINKED ISSUE #421: Possible memory leak of the charset `C.CString` in `(*drv).initCommonCreateParams`
# Possible memory leak of the charset `C.CString` in `(*drv).initCommonCreateParams`

When a charset is configured, `initCommonCreateParams` replaces the package-level encoding
string with a fresh `C.CString` and nothing frees it. Every pool or standalone connection
created with a non-empty `Charset` leaks it.

drv.go:588

```go
// UTF-8 is a shortcut name for AL32UTF8 in ODPI-C (and not the same as the botched UTF8).
var cUTF8, cDriverName = C.CString("UTF-8"), C.CString(DriverName)

// initCommonCreateParams initializes ODPI-C common creation parameters used for creating pools and
// standalone connections. The C strings for the encoding and driver name are
// defined at the package level for convenience.
```

The comment states the invariant the function relies on: the encoding string is a
package-level allocation, so it never needs releasing. drv.go:604 breaks it:

```go
	// assign encoding and national encoding
	P.encoding, P.nencoding = cUTF8, cUTF8
	if charset != "" {
		P.encoding = C.CString(charset)
		P.nencoding = P.encoding
	}
```

`P` is a caller-owned `C.dpiCommonCreateParams`. Neither caller frees `P.encoding`.
drv.go:784, in `acquireConn`:

```go
		if err := d.initCommonCreateParams(&commonCreateParams,
			P.EnableEvents, P.StmtCacheSize,
			P.Charset, P.Token, P.PrivateKey, accessToken,
		); err != nil {
			return nil, false, nil, err
		}
		commonCreateParamsPtr = &commonCreateParams
	}
	// manage strings
	var cUsername, cPassword, cNewPassword, cConnectString, cConnClass *C.char
	defer func() {
		if cUsername != nil {
			C.free(unsafe.Pointer(cUsername))
		}
```

The deferred block covers the five credential strings and not `commonCreateParams.encoding`.
The second caller, `createPool` at drv.go:1097, does the same. The token path in the same
function is released, through `defer freeAccessToken(accessToken)` at drv.go:782, so the
encoding is the one allocation left over.

ODPI-C copies the encoding into the environment it creates rather than keeping the
pointer, so the caller owns the string.

Reaching it needs `Charset` to be set in the connection string, which is not the default;
with the default the code uses the package-level `cUTF8` and there is nothing to free.
For a standalone (non-pooled) configuration this runs once per connection.

Fix, keeping the package-level fast path:

```go
	P.encoding, P.nencoding = cUTF8, cUTF8
	if charset != "" {
		P.encoding = C.CString(charset)
		P.nencoding = P.encoding
	}
```

and have each caller release it when it is not `cUTF8`, alongside the credential strings
they already free.

If you could credit me as a reporter for my contributions to security advisory I will be thankful.

isauran:
### Summary
Adds a regression test for #421 covering `commonCreateParams` with a non-default `Charset`.

Following commits `f55f25a` and `6a80769` (`freeCommonCreateParams`), this test verifies that creating pools / standalone connectors with a custom `Charset` correctly allocates and releases `commonCreateParams.encoding` without leaking native C-heap across repeated iterations.

### Empirical Benchmark & Verification
We ran an empirical stress-test (1,000 iterations allocating and initializing connection parameters with custom charset):
- **Before commit `6a80769` (without `freeCommonCreateParams` in `createPool`)**:
  - Start RSS: 9.19 MB
  - End RSS: 72.31 MB
  - Delta: **+63.12 MB leaked** in unmanaged C-heap (Go GC Alloc flat at 0.22 MB due to CGO blindspot).
- **After commit `6a80769` (with `freeCommonCreateParams` applied)**:
  - Start RSS: 72.50 MB
  - End RSS: 72.64 MB
  - Delta: **+0.14 MB** (flat memory curve, leak fully eliminated).

Tested on Go 1.24 / Go 1.26 (linux/amd64, darwin/arm64).

Fixes #421


tgulacsi (approved):
RSS measurement is missing, NewConnector is created each round...

